PR #2, opened 2018-07-19
Refactor/risk factor

michaella21:
I haven't made a change in tests since we may need to discuss how/what to change for both tests in `test_distributions` and `test_base_risk` which are all pretty messy. Maybe you can review this PR first and we can discuss about tests?


vivarium_public_health/risks/distributions.py
```diff
@@ -94,7 +96,7 @@ def process(self, data: Union[np.ndarray, pd.Series], process_type: str,
         """
         return data
 
-    def pdf(self, x: pd.Series, interpolation: bool=True) -> Union[np.ndarray, pd.Series]:
+    def pdf(self, x: pd.Series, interpolation: bool = True) -> Union[np.ndarray, pd.Series]:
```
collijk: There shouldn't be spaces around the `=` here since it's a kwarg.

vivarium_public_health/risks/distributions.py
```diff
@@ -384,9 +385,46 @@ def ppf(self, x: pd.Series, interpolation: bool = True) -> Union[np.ndarray, pd.
...
+    def __init__(self, exposure_data: pd.DataFrame, risk: str):
+        self.exposure_data = exposure_data
+        self._risk = risk
+
+    def setup(self, builder):
+        self.exposure = builder.value.register_value_producer(f'{self._risk}.exposure',
+                                                              source=builder.lookup.build_table(self.exposure_data))
+    @staticmethod
+    def _get_sorted_categories(data):
+        import re
```
collijk: You usually only put imports inside classes/functions if you're trying to avoid cyclic imports or you're loading a super large dependency that you don't always need.  `re` is in the standard library, so put this import at the top.

vivarium_public_health/risks/distributions.py
```diff
@@ -384,9 +385,46 @@ def ppf(self, x: pd.Series, interpolation: bool = True) -> Union[np.ndarray, pd.
...
+        new_keys = lambda key: [int(elem) if elem.isdigit() else elem.lower() for elem in re.split('([0-9]+)', key)]
```
collijk: Named lambda's almost always mean you're doing something too complicated for a lambda function.  This seems like the case here.

vivarium_public_health/risks/distributions.py
```diff
@@ -384,9 +385,46 @@ def ppf(self, x: pd.Series, interpolation: bool = True) -> Union[np.ndarray, pd.
...
+        return sorted(data, key=new_keys)
+
+    def ppf(self, x):
+        exposure = self.exposure(x.index)
+        # Get a list of sorted category names (e.g. ['cat1', 'cat2', ..., 'cat9', 'cat10', ...])
+        categories = self._get_sorted_categories([column for column in exposure if 'cat' in column])
+        sorted_exposures = exposure[categories]
+        try:
```
collijk: This is a bad pattern.  Instead, try:
```
if not np.allclose(1, np.sum(sorted_exposures, axis=1)):
    raise MissingDataError(...)
```

vivarium_public_health/risks/distributions.py
```diff
@@ -384,9 +385,46 @@ def ppf(self, x: pd.Series, interpolation: bool = True) -> Union[np.ndarray, pd.
...
+            raise MissingDataError('All exposure data returned as 0.')
+        exposure_sum = sorted_exposures.cumsum(axis='columns')
+        category_index = (exposure_sum.T < x).T.sum('columns')
+        return pd.Series(np.array(categories)[category_index], name=self._risk + '_exposure', index=x.index)
+
+
+def get_distribution(risk: str, risk_type: str, builder) -> Union[BaseDistribution, EnsembleDistribution, CategoricalDistribution]:
 
     distribution = builder.data.load(f"{risk_type}.{risk}.distribution")
+    if distribution == "dichotomous" or distribution == "polytomous":
```
collijk: This is totally fine, but I tend to prefer something like `if distribution in ['dichotomous', 'polytomous']:` since it's easier to add additional cases to.

vivarium_public_health/risks/base_risk.py
```diff
@@ -1,47 +1,40 @@
...
-        The name of a risk factor
-    distribution_loader : callable
-        A function which take a builder and returns a standard CEAM
-        lookup table which returns distribution data.
-    exposure_function : callable
-        A function which takes the output of the lookup table created
-        by distribution_loader and a propensity value for each simulant
-        and returns the current exposure to this risk factor.
-    """
+class RiskComponent:
```
collijk: I think we just call this `Risk`.  The `Component` part is redundant.

vivarium_public_health/risks/base_risk.py
```diff
@@ -1,47 +1,40 @@
...
     def setup(self, builder):
-
         self.propensity_function = uncorrelated_propensity
 
         self.exposure_distribution = get_distribution(self._risk, self._risk_type, builder)
         builder.components.add_components([self._effects, self.exposure_distribution])
         self.randomness = builder.randomness.get_stream(self._risk)
         self.population_view = builder.population.get_view(
-            [self._risk+'_exposure', self._risk+'_propensity', 'age', 'sex'])
+            [self._risk + '_exposure', self._risk + '_propensity', 'age', 'sex'])
```
collijk: Good for fixing old style issues.  What do you think about making stuff like this into f-strings?  E.g. `f'{self._risk}_exposure'` instead of `self._risk + '_exposure'`.

vivarium_public_health/risks/base_risk.py
```diff
@@ -43,8 +36,8 @@ def setup(self, builder):
         builder.event.register_listener('time_step__prepare', self.update_exposure, priority=8)
 
     def load_population_columns(self, pop_data):
-        population = self.population_view.get(pop_data.index, omit_missing_columns=True)
-        propensities = pd.Series(self.propensity_function(population, self._risk),
+        population = self.population_view.get(pop_data.index, omit_missing_columns=True )
```
michaella21: When I remove the `omit_missing_columnns=True`, actually the current `test_base_risk` fails for all cases

PR #4, opened 2018-07-25
move exposure/propensity from state table and attach to the pipelines

michaella21:


michaella21:
not yet ready for review, just check point. it does pass `tests/risks/.. ` but need to check other downstream

collijk:
@michaella21 is there a reason to keep this open?  You're more recent risk/indirect effect PRs do most of this.

michaella21:
nope, just forget about it. I will close it now

PR #5, opened 2018-07-26
Remove epidemiology measures

collijk:


michaella21:
I guess deleting the codes are always good(?) but is it not at all used now?

PR #7, opened 2018-08-01
Raise the error when the artifact tries to load the empty data, and a…

michaella21:
…dd the test


michaella21:
ok now it throws some error when running a simulation. I need to fix this first

PR #9, opened 2018-08-02
Allow handling of locations with apostrophes

collijk:


tests/test_dataset_manager.py
```diff
@@ -1,10 +1,19 @@
-import pytest
-
-from vivarium_public_health.dataset_manager import get_location_term
+from vivarium_public_health.dataset_manager import get_location_term, Artifact
```
michaella21: ugh, you deleted pytest

PR #10, opened 2018-08-08
Refactor/artifact entity path

michaella21:


src/vivarium_public_health/dataset_manager.py
```diff
@@ -28,6 +28,27 @@ def parse_artifact_path_config(config):
     return artifact_path
 
 
+class EntityKey(str):
+    def __init__(self, entity_key):
+        self._key = entity_key
```
collijk: You actually don't need to hold on to this at all.  `self` is `entity_key`

src/vivarium_public_health/dataset_manager.py
```diff
@@ -28,6 +28,27 @@ def parse_artifact_path_config(config):
...
+        self.type, self.name, self.measure = self._get_entity_elements()
+        self.group_prefix, self.group_name = self._get_prefix()
+
+    def _get_entity_elements(self):
+        _key_list = self.split('.')
+        name = _key_list[1] if len(_key_list) == 3 else ''
+        return _key_list[0], name, _key_list[-1]
+
+    def _get_prefix(self):
```
michaella21: I probably need to rename this, since it's not really getting prefix

src/vivarium_public_health/dataset_manager.py
collijk: What should happen is you do `super().__init__(entity_key)`

PR #11, opened 2018-08-08
Feature/hdf access patterns

halfhorst:
Fixed some convoluted HDFStore access patterns. Couldn't eliminate all "private" calls, nodes don't actually have public methods.

### DIFF
--- a/src/vivarium_public_health/dataset_manager.py
+++ b/src/vivarium_public_health/dataset_manager.py
@@ -77,7 +77,7 @@ class Artifact:
         return self._cache[cache_key]
 
     def _uncached_load(self, entity_path, keep_age_group_edges, column_filters):
-        group = '/'+entity_path.replace('.', '/')
+        group = entity_path.replace('.', '/')
 
         if group not in self._hdf:
             raise ArtifactException(f"{group} should be in {self.artifact_path}")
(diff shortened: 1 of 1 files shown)


src/vivarium_public_health/dataset_manager.py
```diff
@@ -112,11 +112,11 @@ def _uncached_load(self, entity_key, keep_age_group_edges, column_filters):
         if entity_key.to_path() not in self._hdf:
             raise ArtifactException(f"{entity_key.to_path()} should be in {self.artifact_path}")
 
-        node = self._hdf._handle.get_node(entity_key.to_path())
+        node = self._hdf.get_node(entity_key.to_path())
 
         if "NODE_TYPE" in dir(node._v_attrs) and node.get_attr("NODE_TYPE") == "file":
```
collijk: I'm pretty sure we can use a different mechanism for this same thing without accessing private attributes.  I.e. a quick inspection of one of the artifact hdf files shows that the json blobs are always stored as `tables.Leaf` instances while the data frames are stored as groups (since they are collections of arrays).

halfhorst: Interesting. I was never actually able to find the json in the example artifact so I couldn't play with it. Sounds like this can be reworked

michaella21: If you add any risk factor, its distribution or restrictions comes in as json

PR #12, opened 2018-08-11
Feature/shift write to artifact

collijk:
- Turned the dataset_manager into a package containing the `vivarium` plugin in one module and the artifact in another.
- Pulled in data writing functions into the data artifact.
- Set default filters on the artifact for location and draw, since on loading, these are the dimensions the data grows fastest along.
- pushed all other filtering up into the artifact manager, which makes caching simpler and should increase utilization.
- Built `MockArtifactManager` so that our mocks parallel the classes they're mocking
- Updated tests.


src/vivarium_public_health/dataset_manager/artifact.py
```diff
@@ -0,0 +1,174 @@
...
+        return self.group_prefix + '/' + self.group_name if self.name else self.group_prefix + self.group_name
+
+    def with_measure(self, measure: str) -> 'EntityKey':
+        if self.name:
+            return EntityKey(f'{self.type}.{self.name}.{measure}')
+        else:
+            return EntityKey(f'{self.type}.{measure}')
+
+
+    def to_path(self, measure: str=None) -> str:
```
collijk: Remove measure arg

src/vivarium_public_health/dataset_manager/artifact.py
```diff
@@ -113,8 +111,16 @@ def summary(self):
         return result.getvalue()
 
 
+def get_node_name(node: tables.node.Node):
```
collijk: Remove this

src/vivarium_public_health/dataset_manager/artifact.py
collijk: Except for the leading forward slash, yes.

PR #14, opened 2018-08-24
Hotfix/build artifact default

halfhorst:
This maintains the append error if the user wants to append to a non-existent file, but it removes the file if it exists and append is false.

### DIFF
--- a/src/vivarium_public_health/dataset_manager/hdf.py
+++ b/src/vivarium_public_health/dataset_manager/hdf.py
@@ -44,7 +44,8 @@ def touch(path: str, append: bool):
         if append:
             raise FileNotFoundError("You indicated you want to append to "
                                     f"an existing artifact at {path} but no "
-                                    "such artifact exists.")
+                                    "such artifact exists. remove --append if "
+                                    "you wish to create a new artifact.")
         else:
             with tables.open_file(str(path), mode='w'):
                 pass
--- a/src/vivarium_public_health/dataset_manager/hdf.py
+++ b/src/vivarium_public_health/dataset_manager/hdf.py
@@ -38,7 +38,6 @@ def touch(path: str, append: bool):
     if path.is_file() and append:
         pass
     if path.is_file() and not append:
-        path.unlink()
         with tables.open_file(str(path), mode='w'):
             pass
     elif not path.is_file() and append:
--- a/src/vivarium_public_health/dataset_manager/hdf.py
+++ b/src/vivarium_public_health/dataset_manager/hdf.py
@@ -17,9 +17,9 @@ DEFAULT_COLUMNS = {"year", "location", "draw", "cause", "risk"}
 def touch(path: str, append: bool):
     """Creates an hdf file or wipes an existing file if necessary.
 
-    If overwrite is false and the hdf file exists, touch will do nothing.
-    If the file does not exist, touch will create it. If overwrite is true
-    and a file exists, it will be destroyed and re-created empty.
+    If the file exists and append is not true, the existing file
+    will be destroyed. If the file exists and append is true, touch
+    will do nothing.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


src/vivarium_public_health/dataset_manager/hdf.py
```diff
@@ -15,28 +15,41 @@
...
     FileNotFoundError
         If attempting to append to a non-existent file."""
+
     path = Path(path)
 
-    if append and not path.is_file():
-        raise FileNotFoundError("You indicated you want to append to an existing artifact "
-                                f"at {path} but no such artifact exists.")
-    elif not append:
+    if path.is_file() and not append:
```
collijk: Flatten your cases here.  We have four possible options.  based on `append` and `path.exists()`.  Enumerate them independently.

src/vivarium_public_health/dataset_manager/hdf.py
```diff
@@ -15,28 +15,41 @@
...
+        path.unlink()
```
collijk: Why do you need to unlink the path?  Shouldn't opening the file in write mode truncate it?

halfhorst: It does, I out-thought myself!

src/vivarium_public_health/dataset_manager/hdf.py
```diff
@@ -15,25 +15,37 @@
...
+
     path = Path(path)
 
-    if append and not path.is_file():
-        raise FileNotFoundError("You indicated you want to append to an existing artifact "
-                                f"at {path} but no such artifact exists.")
-    elif not append:
+    if path.is_file() and append:
+        pass
+    if path.is_file() and not append:
```
kfwilson: Why do you need this branch if it does the same thing as the else?

src/vivarium_public_health/dataset_manager/hdf.py
```diff
@@ -15,25 +15,37 @@
...
+        with tables.open_file(str(path), mode='w'):
+            pass
+    elif not path.is_file() and append:
+        raise FileNotFoundError("You indicated you want to append to "
+                                f"an existing artifact at {path} but no "
+                                "such artifact exists. remove --append if "
```
kfwilson: I'm not familiar enough with how this gets used to know, but is it only ever called by your build_artifact script? If not, the reference to --append might be confusing

src/vivarium_public_health/dataset_manager/hdf.py
halfhorst: The else will wipe the file by opening in write mode, is_file() and append actually does nothing. I included it just to enumerate all the branches for clarity

src/vivarium_public_health/dataset_manager/hdf.py
halfhorst: Well, the append comes here from the config which is set in the build_artifact script, and you can hit this error as a result of calling build_artifact (this is the spot where I was getting the original error that started these PRs)

halfhorst: But, I don't know if touch would be called somewhere else that would then make this message confusing

src/vivarium_public_health/dataset_manager/hdf.py
kfwilson: I'm talking about the is_file and not append branch though - that seems like it also wipes the file?

src/vivarium_public_health/dataset_manager/hdf.py
halfhorst: Easiest would just be to remove the extra sentence. We could catch this exception in build artifact and add the message about the append flag, I suppose

kfwilson: Yeah I guess saying "you indicated you want to append" does make it clear in the context of build_artifact that you're referring to the --append flag and it's more general

PR #15, opened 2018-08-29
Bugfix/filter columns hdf load

kfwilson:
This is a workable (but not necessarily pretty) fix to the fact that the base filter terms (which reference draws and location) were applied to all keys in the data artifact, which broke things like population.structure that don't have draws. Now any filter terms referencing columns that aren't present in the data are skipped.


kfwilson:
Added unit tests for _get_valid_filter_terms and then tests for artifact load (to ensure that the artifact's filter terms don't change even if the full set isn't valid for an individual key) and for hdf load with both valid and invalid filters

PR #20, opened 2018-10-06
resolve conflicts of population size default configurations

michaella21:
just change some values in component and tests according to the change in  previous PR in vivarium

### DIFF
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -18,7 +18,7 @@ class BasePopulation:
             'age_start': 0,
             'age_end': 125,
             'exit_age': None,
-            'population_size': 10000,
+            'population_size': 100,
         }
     }
 
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -18,7 +18,6 @@ class BasePopulation:
             'age_start': 0,
             'age_end': 125,
             'exit_age': None,
-            'population_size': 100,
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/vivarium_public_health/population/base_population.py
```diff
@@ -18,7 +18,7 @@ class BasePopulation:
             'age_start': 0,
             'age_end': 125,
             'exit_age': None,
-            'population_size': 10000,
+            'population_size': 100,
```
collijk: This should still fail, I think.  `population_size` is set by the population manager in vivarium.

tests/disease/test_disease.py
```diff
@@ -52,7 +52,7 @@ def test_dwell_time(assign_cause_mock, base_config, disease, base_data):
     base_config.update({
         'time': {'step_size': time_step},
         'population': {'population_size': 10}
-    }, **metadata(__file__))
+    }, layer='override')
```
collijk: Why the change here?

src/vivarium_public_health/population/base_population.py
michaella21: when I ran the pytest before this change, test_base_population failed with the population size 100 set by population manager that was still a component level configs, then base_population tries to override at the same config level.

tests/disease/test_disease.py
michaella21: probably i didn't think enough :(

halfhorst:
So this just eliminates a setting of population_size because it would break the "multiple components shouldn't set defaults for the same thing" PR ?

PR #22, opened 2018-10-18
Eliminate global state in mock artifact and fix tests

halfhorst:
Eliminate accidental global state in the mock artifact, and correct the tests that it breaks.

Please take a look at the test changes, they were simple but I want to make sure they don't render the tests vacuous

### DIFF
--- a/tests/risks/test_effect.py
+++ b/tests/risks/test_effect.py
@@ -279,9 +279,9 @@ def test_CategoricalRiskComponent_dichotomous_case(base_config, base_plugins):
 
     simulation.data.write("risk_factor.test_risk.relative_risk", rr_data)
     simulation.data.write("risk_factor.test_risk.population_attributable_fraction", 1)
-
     simulation.data.write("risk_factor.test_risk.affected_causes", affected_causes)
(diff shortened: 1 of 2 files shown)


tests/risks/test_effect.py
```diff
@@ -279,9 +279,9 @@ def test_CategoricalRiskComponent_dichotomous_case(base_config, base_plugins):
 
     simulation.data.write("risk_factor.test_risk.relative_risk", rr_data)
     simulation.data.write("risk_factor.test_risk.population_attributable_fraction", 1)
-
     simulation.data.write("risk_factor.test_risk.affected_causes", affected_causes)
     simulation.data.write("risk_factor.test_risk.distribution", "dichotomous")
+    simulation.data.write("risk_factor.test_risk.affected_risk_factors", [])
```
michaella21: yep, I think we only required only `affectd_causes` for risk and `affected_causes` and `affected_risk_factors` for coverage_gap before James rewrote the risk_effect. Then it changes to require both from risk also. good catch.

collijk:
We don't want to approve changes until tests pass against the appropriate upstream branch.

PR #23, opened 2018-10-19
remove list/tuple from default configuration

michaella21:
small PR for [removing lists/tuples from the mass treatment campaign configuration](https://jira.ihme.washington.edu/browse/MIC-246 ). The corresponding PR for ceam_experiments is also coming.


michaella21:
and it fails from the tests requiring the recent change, that hasn't been merged to develop.

PR #25, opened 2018-10-24
Feature/extend interactive context

kfwilson:
Adding interactive initialize/setup sim methods to vivarium public health that by default set up the artifact manager

### DIFF
--- a/src/vivarium_public_health/interactive.py
+++ b/src/vivarium_public_health/interactive.py
@@ -1,7 +1,9 @@
-from vivarium.interface import interactive
-from vivarium.framework.configuration import build_simulation_configuration, build_model_specification
+import vivarium.interface.interactive as vi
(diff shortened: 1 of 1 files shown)


src/vivarium_public_health/interactive.py
```diff
@@ -0,0 +1,54 @@
...
+
+def setup_simulation(components: List, input_config: Union[Mapping, ConfigTree]=None,
+                     plugin_config: Mapping=None) -> vi.InteractiveContext:
+    simulation = initialize_simulation(components, input_config, plugin_config)
+    simulation.setup()
+
+    return simulation
+
+
+def initialize_simulation_from_model_specification(model_specification_file: str) -> vi.InteractiveContext:
```
collijk: Here as well.  Drop line 40 and replace the explicit constructions with a call to the locally defined `initialize_simulation`

PR #26, opened 2018-10-24
Feature/expose prevalence propensity

kfwilson:
exposing prevalence propensity in pipeline + switching assignment of initial states to use that pipeline (instead of choice in randomness)

### DIFF
--- a/src/vivarium_public_health/disease/model.py
+++ b/src/vivarium_public_health/disease/model.py
@@ -73,24 +73,31 @@ class DiseaseModel(Machine):
     def get_csmr(self):
         return self._csmr_data
 
-    def get_state_map(self, pop_index):
-        return
+    def get_state_weights(self, pop_index):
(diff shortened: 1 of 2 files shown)


src/vivarium_public_health/disease/model.py
```diff
@@ -70,20 +74,30 @@ def get_csmr(self):
...
         simulants = simulants_df[['age', 'sex']].copy()
         sequelae, weights = zip(*states.items())
         sequelae += (initial_state,)
         for w in weights:
             w.reset_index(inplace=True, drop=True)
         weights += ((1 - np.sum(weights, axis=0)),)
-        simulants.loc[:, 'condition_state'] = randomness.choice(simulants.index, sequelae,
-                                                                np.array(weights).T)
+
+        # manually calculate the assigned initial states in order to use propensities
```
collijk: I think this function is now probably doing too much work.  Maybe we can make a `self.get_state_map` function that gets 
called on line 103 instead of the dict comprehension that handles a lot of this weight normalization up front.  Up to you.

PR #27, opened 2018-10-26
Feature/rebin polytomous risk

michaella21:
This PR is to rebin the Polytomous risk exposure and RR so that we can have the indirect effects as Dichotomous risk factor. It is validated and tested but still need to add tests at least in `effects`. I'm making a PR now since everyone wants to run their model with this component. To use the rebin effects, as we discussed in the design meeting, you need to specify in the configuration with the `{risk_name}: rebin: True`. I will add more tests while it's reviewed if there's no more rush for research support.

### DIFF
--- a/src/vivarium_public_health/risks/base_risk.py
+++ b/src/vivarium_public_health/risks/base_risk.py
@@ -17,6 +17,7 @@ class Risk:
        risk_name : str
            The name of a risk
        """
+    configuration_defaults = {}
 
     def __init__(self, risk_type, risk_name):
         self._risk_type, self._risk = risk_type, risk_name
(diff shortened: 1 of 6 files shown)


michaella21:
Now actually ready for review. In test_distributions, `test_distributions` was pretty bad earlier. I **attempted** to fix it but didn't work out. It is still pretty messy. Individual distribution test was not touchable since it was for checking whether our math is correct. Last 3 test in `test_distributions` have lots of duplicated codes but I wasn't really sure whether it's necessary to refactor. Maybe it is :(

collijk:
Move data munging code to separate module then merge.

PR #29, opened 2018-11-01
Refactor/distributions

kfwilson:
Changes to accompany pulling out distributions into separate package

Tests fail because it doesn't have the newest separate distributions package yet but passes locally

All smoke tests also pass with these changes


src/vivarium_public_health/risks/distributions.py
```diff
@@ -1,403 +1,61 @@
...
-            except NonConvergenceError as e:
-                if weights[e.dist] > 0.05:
-                    weights = weights.drop(e.dist)
-                else:
-                    raise NonConvergenceError(f'Divergent {key} distribution has weights: {100*weights[key]}%', key)
+        builder.components.add_components(self._distributions.values())
 
-        return weights/np.sum(weights), dist
+    @classmethod
+    def get_distribution_map(cls):
```
collijk: We don't want class methods for `get_distribution_map` (and therefore on `get_valid_distribution`).  With normal inheritance this method looks like:

```
def get_distribution_map(self):
    dist_map = super().get_distribution_map()
    return {dist_name: partial(SimulationDistribution, distribution=dist) for dist_name, dist in dist_map.items()}
```

PR #30, opened 2018-11-02
Refactor/paf by risk

michaella21:
We decided to change the way of pulling paf (by cause -> by risk) According to the changes in data pulling and loading, we also compute all the pafs for the categorical risk (and coverage gap) on the fly. Many of changes are just moving around the data transformation code and make the tests work along with the new changes.

### DIFF
--- a/tests/risks/test_data_transformation.py
+++ b/tests/risks/test_data_transformation.py
@@ -2,89 +2,67 @@ import pandas as pd
 import numpy as np
 
 import pytest
+from typing import List, Union
 
 from vivarium.testing_utilities import build_table
-from vivarium.framework.configuration import build_simulation_configuration
 from vivarium_public_health.risks.data_transformation import *
 
-def test_should_rebin():
-    test_config = build_simulation_configuration()
-    test_config['population'] = {'population_size': 100}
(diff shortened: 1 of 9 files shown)


tests/risks/test_data_transformation.py
```diff
@@ -0,0 +1,117 @@
...
+    test_config['test_risk'] = {}
+    assert not should_rebin('test_risk', test_config)
+
+    test_config['test_risk'].rebin = False
+    assert not should_rebin('test_risk', test_config)
+
+    test_config['test_risk']['rebin'] = True
+    assert should_rebin('test_risk', test_config)
+
+
```
collijk: We generally need to take better advantage of pytest fixtures and parameterization.  A lot of tests look like this where we're mixing the generation of test data with the logic of the test.  That makes them very hard to maintain in the long run.

michaella21:
I took out the data generation  for each type of risk from individual tests and make them as fixtures. But most tests in `test_effect` are integration tests and thus they are very slow to run also. But I do not have a good idea how to make it more efficient and make it as actual unit tests for now. We probably need to revisit it again.

PR #31, opened 2018-11-02
Feature/future open cohort

michaella21:
Added a constant extrapolation,  Whenever simulation requires a birth rate beyond the year available, population will constantly grow at the same rate of whenever (currently 2016) the last year of available data

### DIFF
--- a/tests/population/test_add_new_birth_cohort.py
+++ b/tests/population/test_add_new_birth_cohort.py
@@ -1,3 +1,4 @@
+import pytest
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


tests/population/test_add_new_birth_cohort.py
```diff
@@ -65,6 +64,50 @@ def test_FertilityCrudeBirthRate(base_config, base_plugins):
     # TODO: Write a more rigorous test.
     assert len(pop.age) > start_population_size, 'expect new simulants'
 
+    base_config.update({
```
collijk: We want to try not to add more integration tests like this if we can avoid it.  Maybe have ``add_new_birth_cohort`` method pass the population table into the ``_get_birth_rate`` function and make that function static.  Then you can write a proper unit test that doesn't involve setting up and running a simulation.

tests/population/test_add_new_birth_cohort.py
```diff
@@ -65,6 +64,50 @@ def test_FertilityCrudeBirthRate(base_config, base_plugins):
...
+
+    simulation = initialize_simulation(components, base_config, base_plugins)
+    simulation.data.write("covariate.live_births_by_sex.estimate",
+                          build_table(500, 1990, 2016, ('age', 'year', 'sex', 'mean_value')
+                                      ).query('age == 25').drop('age', 'columns'))
+    simulation.setup()
+    with pytest.raises(ValueError):
+        simulation.take_steps(6)
+
+    base_config.update({
```
collijk: Either parameterize the test with configuration/data or write two separate tests for the different set ups.

PR #33, opened 2018-11-08
Bugfix/generate population in between data years

kfwilson:
This is a bug I noticed when I was going through to pull out age and year and decided to fix it separately so the logic changes could be reviewed without the mess of all the age/year drops. 

The bug was that previously, unless we had population structure data for every year, it would use the first year of data in generating a base population. This isn't a problem in all of the actual simulations because gbd does have population data for every year. But it was a quick bug to fix, and in a lot of our tests for other parts, we use made up data with years like [1990, 1995, 2000, 2005]. In that case, asking for the population data to use for 1997 would give you the data for 1990 (the first and second commits in this PR were to show that was how it was behaving previously).

### DIFF
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -1,4 +1,5 @@
 import pandas as pd
+import numpy as np
 
 from .data_transformations import assign_demographic_proportions, rescale_binned_proportions, smooth_ages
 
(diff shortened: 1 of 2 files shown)


src/vivarium_public_health/population/base_population.py
```diff
@@ -44,6 +45,11 @@ def setup(self, builder):
         if self.config.exit_age is not None:
             builder.components.add_components([AgedOutSimulants()])
 
+    @staticmethod
+    def select_sub_population_data(reference_population_data, year):
+        reference_years = sorted(set(reference_population_data.year))
+        ref_year_index = np.digitize(year, reference_years).item()-1
```
halfhorst: digitize, how cool :grin:

halfhorst:
neat, good catch!

PR #35, opened 2018-11-15
Feature/artifact filter via config

kfwilson:
Make filter_terms on artifact private to indicate they shouldn't be changed once an artifact is created (otherwise you get unexpected results on load b/c of the cache).

Add the ability to specify a single filter term in the config under 'artifact_filter_term' which is applied to all data containing the column specified in the filter term. 

Corresponding PR in vivarium_inputs for passthrough. 

Added unit tests + tested by specifying a filter term in config restricting years and then trying to run sim outside those years w/o extrapolation with artifact to make sure it fails.

### DIFF
--- a/src/vivarium_public_health/dataset_manager/artifact.py
+++ b/src/vivarium_public_health/dataset_manager/artifact.py
@@ -42,6 +42,10 @@ class Artifact:
         """A list of all the keys contained within the artifact."""
         return self._keys
 
+    @property
+    def filter_terms(self) -> List[str]:
+        return self._filter_terms
+
     def load(self, entity_key: str) -> Any:
         """Loads the data associated with provided EntityKey.
 
--- a/src/vivarium_public_health/dataset_manager/artifact.py
+++ b/src/vivarium_public_health/dataset_manager/artifact.py
@@ -32,7 +32,7 @@ class Artifact:
         """
 
         self.path = path
-        self.filter_terms = filter_terms
+        self._filter_terms = filter_terms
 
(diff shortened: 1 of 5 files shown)


src/vivarium_public_health/dataset_manager/artifact.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, path: str, filter_terms: List[str]=None):
         """
 
         self.path = path
-        self.filter_terms = filter_terms
+        self._filter_terms = filter_terms
```
collijk: Make into a property with no setter.

src/vivarium_public_health/dataset_manager/dataset_manager.py
```diff
@@ -105,16 +112,33 @@ def load(self, entity_key: str, keep_age_group_edges: bool=True, **column_filter
             if the data is a dataframe.
         """
         data = self.artifact.load(entity_key)
-        return filter_data(data, keep_age_group_edges, **column_filters) if isinstance(data, pd.DataFrame) else data
+        return filter_data(data, keep_age_group_edges, self.config_filter_term, **column_filters) if isinstance(data, pd.DataFrame) else data
```
halfhorst: Is it possible for filter_data to return anything other than a DF?

collijk: Filter data will break if it gets something that is not a dataframe and will always return a dataframe if it receives one.

PR #37, opened 2018-11-19
Feature/artifact replace

kfwilson:
adding replace option for artifact + tests

### DIFF
--- a/tests/dataset_manager/test_artifact.py
+++ b/tests/dataset_manager/test_artifact.py
@@ -1,5 +1,5 @@
 import pytest
-import pandas as pd
+from unittest.mock import call
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


tests/dataset_manager/test_artifact.py
```diff
@@ -264,6 +264,39 @@ def test_loading_key_leaves_filters_unchanged():
...
+
+    a = Artifact(path, filter_terms)
+
+    assert ekey not in a.keys
+
+    a.write(key, "data")
+
+    a.replace(key, "new_data")
+    hdf_mock.remove.assert_called_with(path, ekey)
+    hdf_mock.write.assert_called_with(path, ekey, "new_data")
```
collijk: You should check for all calls: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_has_calls

PR #38, opened 2018-11-26
Bugfix/fertility rate filter

kfwilson:
Hit some bugs in the existing FertilityAgeSpecificRates component when I was trying to use it for the forecasting data where it was trying to pull a column that didn't exist and wasn't filtering down to just women so ended up with duplicates that broke in interpolation.

### DIFF
--- a/src/vivarium_public_health/population/add_new_birth_cohorts.py
+++ b/src/vivarium_public_health/population/add_new_birth_cohorts.py
@@ -178,7 +178,7 @@ class FertilityAgeSpecificRates:
         self.randomness = builder.randomness.get_stream('fertility')
         asfr_data = builder.data.load("covariate.age_specific_fertility_rate.estimate")[['year_start', 'year_end',
                                                                                          'age_group_start',
-                                                                                         'age_group_end', 'value']]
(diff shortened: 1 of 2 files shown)


src/vivarium_public_health/population/add_new_birth_cohorts.py
```diff
@@ -176,9 +176,9 @@ def setup(self, builder):
         """
 
         self.randomness = builder.randomness.get_stream('fertility')
-        asfr_data = builder.data.load("covariate.age_specific_fertility_rate.estimate")[['year_start', 'year_end',
-                                                                                         'age_group_start',
-                                                                                         'age_group_end', 'value']]
+        asfr_data = builder.data.load("covariate.age_specific_fertility_rate.estimate")
+        asfr_data = asfr_data[asfr_data.sex == 'Female'][['year_start', 'year_end',
```
kfwilson: GBD includes Males in its fertility rate data, although the values are all 0

michaella21: same here. Many of sex restricted (or age restricted) data often ends up having restricted age/sex with all 0 data :( (sometimes also have non-zero data)

PR #42, opened 2018-12-13
Feature/data free components

kfwilson:
I had to make a clean branch and just copy over all the data free component changes because I totally muddled things with the risk-effect-rewrite branch - once that was merged in, it was overwriting and deleting a lot of these components  

This is just the data free component changes + Cody's fix of the risk/risk effect docstrings

### DIFF
--- a/src/vivarium_public_health/risks/base_risk.py
+++ b/src/vivarium_public_health/risks/base_risk.py
@@ -9,18 +9,26 @@ class Risk:
     (1) high systolic blood pressure as a risk where the SBP is not dichotomized
         into hypotension and normal but is treated as the actual SBP measurement.
     (2) smoking as two categories: current smoker and non-smoker.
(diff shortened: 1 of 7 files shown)


src/vivarium_public_health/risks/base_risk.py
```diff
@@ -36,17 +43,24 @@ def setup(self, builder):
 
     def _get_distribution(self, builder):
         """A wrapper to isolate builder from setup"""
+        kwargs = {}
+        if builder.configuration[self._risk]['exposure'] != 'data':
+            if builder.configuration[self._risk]['distribution'] != "dichotomous":
+                raise ValueError("A Dummy Risk must have a dichotomous distribution.")
```
collijk: Raise a not implemented error here.  We may support more complicated things later.

PR #44, opened 2018-12-18
Feature/cause level disability weights

halfhorst:


halfhorst:
standby, need to address these tests

halfhorst:
this is good now

tests/disease/test_disease.py
```diff
@@ -57,6 +59,7 @@ def test_dwell_time(assign_cause_mock, base_config, disease, base_data):
     healthy_state = BaseDiseaseState('healthy')
     data_function = base_data(0)
     data_function['dwell_time'] = lambda _, __: pd.Timedelta(days=28)
+    data_function['disability_weight'] = lambda _, __: build_table(0.0, year_start-1, year_end)
```
kfwilson: these changes are all also in your other PR so I'm not sure where you want to address things, but I don't think you need to be using build_table - it should work to return a single value that gets passed to builder.lookup.build_table

halfhorst: yup I'm keeping the models up to date with this, it depends on it. I'm altering that now

PR #45, opened 2018-12-18
Feature/morbidity models

halfhorst:
These are factories for producing disease models that formerly lived in ceam_experiments in the conic folder (and still do). I think that some of them are currently WIP, and the core logic was not re-written at all. They have been moved in to vph and also altered to support morbidity only models in addition to models with omrtality. This means overriding the CSMR data function if the model is morbidity only as well as pulling disability weights by cause. 

I tested this with a bunch of simple configurations that use the models, an example of which is attached. The only one I'm not quite sure about is neonatal because I couldn't figure out a good candidate that would have morbidity.

[morbidity_test.txt](https://github.com/ihmeuw/vivarium_public_health/files/2689398/morbidity_test.txt)

### DIFF
--- a/src/vivarium_public_health/disease/models.py
+++ b/src/vivarium_public_health/disease/models.py
@@ -98,7 +98,7 @@ class SIS_fixed_duration:
         """
         self.cause = cause
         if not isinstance(duration, pd.Timedelta):
-            self.duration = pd.Timedelta(days=float(duration) // 1, hours=float(duration) % 1)
+            self.duration = pd.Timedelta(days=float(duration) // 1, hours=(float(duration) % 1) * 24.0)
         else:
             self.duration = duration
 


src/vivarium_public_health/risks/base_risk.py
```diff
@@ -89,3 +89,50 @@ def get_current_exposure(self, index):
 
     def __repr__(self):
         return f"Risk(_risk_type= {self._risk_type}, _risk= {self._risk})"
+
+
+class DummyRisk(Risk):
```
collijk: No bueno

src/vivarium_public_health/population/mortality.py
```diff
@@ -74,6 +74,8 @@ def metrics(self, index, metrics):
         metrics['total_population__dead'] = len(the_dead)
 
         for (condition, count) in pd.value_counts(the_dead.cause_of_death).to_dict().items():
+            if condition == "death_due_to_other_causes":
```
collijk: Why the change?

src/vivarium_public_health/disease/state.py
```diff
@@ -196,17 +196,11 @@ def setup(self, builder):
...
         self._dwell_time = get_dwell_time_func(self.cause, builder)
         self.randomness_prevalence = builder.randomness.get_stream(f'{self.state_id}_prevalent_cases')
 
-        if isinstance(disability_weight_data, pd.DataFrame):
-            self._disability_weight = builder.lookup.build_table(float(disability_weight_data.value))
-        elif disability_weight_data is not None:
-            self._disability_weight = builder.lookup.build_table(disability_weight_data)
-        else:
-            self._disability_weight = builder.lookup.build_table(0)
+        self._disability_weight = builder.lookup.build_table(get_disability_weight_func(self.cause, builder))
```
collijk: Yay!

src/vivarium_public_health/risks/base_risk.py
halfhorst: Yeah I don't know what's goin on here. I'm removing it

src/vivarium_public_health/population/mortality.py
halfhorst: so this is another change I was going to filter out after the fact. I got annoyed that the simulation summary at the end would say "deaths_due_to_death_due_to_other_causes." Now it just says "deaths_due_to_other_causes"

src/vivarium_public_health/disease/state.py
halfhorst: this breaks one unit test because of our default getters :(

src/vivarium_public_health/population/mortality.py
collijk: Sweet!

collijk: Probably not the right place to fix this though

src/vivarium_public_health/disease/state.py
halfhorst: fixed

src/vivarium_public_health/disease/models.py
```diff
@@ -0,0 +1,154 @@
+import pandas as pd
+
+from vivarium_public_health.disease import (SusceptibleState, ExcessMortalityState, RecoveredState,
+                                            DiseaseState, DiseaseModel)
+from vivarium_inputs.utilities import DataMissingError
```
collijk: No way, man.  vph can't import from vivarium inputs.

halfhorst: updated

src/vivarium_public_health/disease/models.py
```diff
@@ -0,0 +1,152 @@
...
+        builder.components.add_components([DiseaseModel(self.cause, states=[healthy, infected],
+                                                        get_data_functions=get_data_functions)])
+
+
+class SIS_fixed_duration:
+
+    def __init__(self, cause: str, duration):
+        self.cause = cause
+        if not isinstance(duration, pd.Timedelta):
+            self.duration = pd.Timedelta(days=float(duration) // 1, hours=float(duration) % 1)
```
kfwilson: Can we add a doc string saying how duration should be specified? This makes it seem like if I did something like duration=28.5, I'd end up with 28 days and 0.5 hours which doesn't really make sense

src/vivarium_public_health/disease/state.py
```diff
@@ -196,17 +196,11 @@ def setup(self, builder):
             'prevalence', lambda cause, builder: builder.data.load(f"{self.cause_type}.{cause}.prevalence"))
         get_dwell_time_func = self._get_data_functions.get('dwell_time', lambda *args, **kwargs: pd.Timedelta(0))
 
-        disability_weight_data = get_disability_weight_func(self.cause, builder)
         self.prevalence_data = builder.lookup.build_table(get_prevalence_func(self.cause, builder))
         self._dwell_time = get_dwell_time_func(self.cause, builder)
         self.randomness_prevalence = builder.randomness.get_stream(f'{self.state_id}_prevalent_cases')
 
-        if isinstance(disability_weight_data, pd.DataFrame):
```
halfhorst: @kfwilson the single value was breaking my tests and I thought it was because of this change. I'll check agian

tests/disease/test_disease.py
```diff
@@ -57,6 +59,7 @@ def test_dwell_time(assign_cause_mock, base_config, disease, base_data):
     healthy_state = BaseDiseaseState('healthy')
     data_function = base_data(0)
     data_function['dwell_time'] = lambda _, __: pd.Timedelta(days=28)
+    data_function['disability_weight'] = lambda _, __: build_table(0.0, year_start-1, year_end)
```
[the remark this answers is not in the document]
michaella21: I think build_table is not the interpolation table, isn't it the one made for the testing? it's in testing utilities I think, just making our common shape dataframe with age,year,sex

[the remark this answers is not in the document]
halfhorst: yeah it looks good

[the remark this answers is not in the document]
kfwilson: build_table isn't the interpolated table but the results of the disability_weight data function get passed to a lookup table which should be able to handle either a single value or a df

src/vivarium_public_health/disease/state.py
kfwilson: weird. I would have thought it would be fine since it gets passed to a lookup table

src/vivarium_public_health/disease/models.py
halfhorst: @collijk can I multiply hourse by 24 ? I agree with kate, this is a little confusing

collijk: Yep, think this is an error.

PR #46, opened 2018-12-19
Bugfix/pipeline not filter untracked simulants

michaella21:
Actual changes are in distributions.py and transition.py to use the scalar table instead of the pd.Series. Other changes are all replacing the sim.population.population to sim.get_population()

### DIFF
--- a/src/vivarium_public_health/disease/transition.py
+++ b/src/vivarium_public_health/disease/transition.py
@@ -14,9 +14,8 @@ class RateTransition(Transition):
         rate_data, pipeline_name = self._get_rate_data(builder)
         self.base_rate = builder.lookup.build_table(rate_data)
         self.effective_rate = builder.value.register_rate_producer(pipeline_name, source=self.rates)
-        self.base_paf = lambda index: [builder.lookup.build_table(0)(index)]
         self.joint_paf = builder.value.register_value_producer(f'{self.output_state.state_id}.incidence_rate.paf',
-                                                               source=self.base_paf,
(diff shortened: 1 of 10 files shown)


src/vivarium_public_health/disease/transition.py
```diff
@@ -14,11 +14,11 @@ def setup(self, builder):
         rate_data, pipeline_name = self._get_rate_data(builder)
         self.base_rate = builder.lookup.build_table(rate_data)
         self.effective_rate = builder.value.register_rate_producer(pipeline_name, source=self.rates)
+        self.base_paf = lambda index: [builder.lookup.build_table(0)(index)]
```
kfwilson: why does this need to be a distinct attribute?

collijk: I don't think it does

tests/validation_tests/rates.py
```diff
@@ -33,23 +33,23 @@ def test_incidence_rate_recalculation(base_config):
...
 
-    susceptible = [(sim.population.population.simple_disease == 'healthy').sum()]
+    susceptible = [(sim.get_population().simple_disease == 'healthy').sum()]
     new_cases = []
     known_cases = pd.Index([])
 
     for step in range(360):
         sim.step()
-        cases = sim.population.population.query('simple_disease == "sick"')
+        cases = sim.population_.population.query('simple_disease == "sick"')
```
collijk: Is this a typo?

PR #51, opened 2018-12-24
Feature/aggregate multil loc artifacts

michaella21:
To support the artifact aggregation, metadata is added to the artifact. Instead of traversing the hdf file to get keys, now it will just read from the `medata.keyspace`.

### DIFF
--- a/src/vivarium_public_health/dataset_manager/artifact.py
+++ b/src/vivarium_public_health/dataset_manager/artifact.py
@@ -101,7 +101,6 @@ class Artifact:
             pass
         else:
             self._keys.append(entity_key)
-
             new_keyspace = self._keys.copy()
             hdf.remove(self.path, EntityKey('metadata.keyspace'))
             hdf.write(self.path, EntityKey('metadata.keyspace'), new_keyspace)
@@ -128,6 +127,7 @@ class Artifact:
         if entity_key in self._cache:
             self._cache.pop(entity_key)
         new_keyspace = self._keys.copy()
+        hdf.remove(self.path, entity_key)
         hdf.remove(self.path, EntityKey('metadata.keyspace'))
         hdf.write(self.path, EntityKey('metadata.keyspace'), new_keyspace)
 
--- a/src/vivarium_public_health/dataset_manager/artifact.py
+++ b/src/vivarium_public_health/dataset_manager/artifact.py
@@ -31,7 +31,7 @@ class Artifact:
             A set of terms suitable for usage with the ``where`` kwarg for ``pd.read_hdf``
         """
 
-        Artifact.create_hdf_with_keyspace(path)
+        self.create_hdf_with_keyspace(path)
         self.path = path
         self._filter_terms = filter_terms
         self._cache = {}
--- a/src/vivarium_public_health/dataset_manager/artifact.py
+++ b/src/vivarium_public_health/dataset_manager/artifact.py
@@ -270,24 +270,36 @@ def _to_tree(keys: List[EntityKey]) -> Dict[str, Dict[str, List[str]]]:
 
 
 class Keys:
+    """A convenient wrapper around the keyspace which makes easier for Artifact
+     to maintain its keyspace when EntityKey is added or removed.
+     With the artifact_path, Keys object is initialized when the Artifact is
+     initialized """
 
     keyspace_node = EntityKey('metadata.keyspace')
 
-    def __init__(self, artifact_path):
+    def __init__(self, artifact_path: str):
         self.artifact_path = artifact_path
         self._keys = [str(k) for k in hdf.load(self.artifact_path, EntityKey('metadata.keyspace'), None)]
 
-    def append(self, new_key):
+    def append(self, new_key: EntityKey):
+        """ Whenever the artifact gets a new key and new data, append is called to
+        remove the old keyspace and to write the updated keyspace"""
+
         self._keys.append(str(new_key))
         hdf.remove(self.artifact_path, self.keyspace_node)
         hdf.write(self.artifact_path, self.keyspace_node, self._keys)
 
-    def remove(self, removing_key):
+    def remove(self, removing_key: EntityKey):
+        """ Whenever the artifact removes a key and data, remove is called to
+        remove the key from keyspace and write the updated keyspace."""
+
         self._keys.remove(str(removing_key))
         hdf.remove(self.artifact_path, self.keyspace_node)
         hdf.write(self.artifact_path, self.keyspace_node, self._keys)
 
-    def to_list(self):
+    def to_list(self) -> List[EntityKey]:
+        """A list of all the EntityKeys in the associated artifact."""
+
         return [EntityKey(k) for k in self._keys]
 
     def __contains__(self, item):
--- a/src/vivarium_public_health/dataset_manager/artifact.py
+++ b/src/vivarium_public_health/dataset_manager/artifact.py
@@ -6,6 +6,7 @@ convenient access and inspection.
 """
 from collections import defaultdict
 import logging
+from pathlib import Path
 from typing import List, Dict, Any
 
 from vivarium_public_health.dataset_manager import hdf
@@ -21,7 +22,6 @@ class ArtifactException(Exception):
 
 class Artifact:
     """An interface for interacting with ``vivarium`` hdf artifacts."""
-
     def __init__(self, path: str, filter_terms: List[str]=None):
         """
         Parameters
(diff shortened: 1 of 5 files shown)


src/vivarium_public_health/dataset_manager/artifact.py
```diff
@@ -93,13 +93,17 @@ def write(self, entity_key: str, data: Any):
...
         """
+
         entity_key = EntityKey(entity_key)
         if entity_key in self.keys:
             raise ArtifactException(f'{entity_key} already in artifact.')
         elif data is None:
             pass
         else:
             self._keys.append(entity_key)
+            new_keyspace = self._keys.copy()
```
kfwilson: I still get a little confused about when you need to copy things in python or not so why are you coping self._keys here and using that instead of just writing self._keys into 'metadata.keyspace'?

collijk: You only need to copy things if you're hanging on to a mutable data structure (lists, dicts, or sets) and you're handing that data structure to a function that might have side effects.  Python doesn't provide great tools for controlling this behavior like you'd get in a language with a more functional flavor.  

That being said, we shouldn't need to make a copy here.  Functions in the `hdf` module should be side effect free w/r/t their input arguments.  If that's not the case, we should look there to make the changes (e.g. have those functions copy their args before mutating).

src/vivarium_public_health/dataset_manager/artifact.py
```diff
@@ -33,9 +34,8 @@ def __init__(self, path: str, filter_terms: List[str]=None):
 
         self.path = path
         self._filter_terms = filter_terms
-
         self._cache = {}
-        self._keys = [EntityKey(k) for k in hdf.get_keys(self.path)]
+        self._keys = [EntityKey(k) for k in hdf.load(path, EntityKey('metadata.keyspace'), None)]
```
collijk: We should think very carefully about encapsulation here.  Before we introduced the `metadata` keys, we didn't actually have hard requirements on what was in the artifact.  This update means we have to create a new hdf file, then also put in an empty list at the hdf node `metadata.keyspace`.  All of that logic is hidden away in the artifact builder code in a separate library.  We should probably consider, I don't know, a class method constructor to generate a new, empty artifact?  Ideas? @michaella21 @kfwilson

src/vivarium_public_health/dataset_manager/artifact.py
michaella21: Actually it was wrong to begin with. Since we all forgot about the fact that we decided to have metadata as strings, not `EntityKey` means, `new_keyspace=[str(k) for k in self._keys]` and yes, no copy

src/vivarium_public_health/dataset_manager/artifact.py
michaella21: yeah, I agree that it has a problem but I also didn't have any good idea how to handle this. There may be some way to handle the location inside of Artifact class but if we want to include versions, (and furthermore the list of components), we need a bridge between builder and Artifact class, maybe a classmethod factory, or any other idea?

src/vivarium_public_health/dataset_manager/artifact.py
```diff
@@ -33,9 +34,16 @@ def __init__(self, path: str, filter_terms: List[str]=None):
 
         self.path = path
         self._filter_terms = filter_terms
-
         self._cache = {}
-        self._keys = [EntityKey(k) for k in hdf.get_keys(self.path)]
+        self._keys = [EntityKey(k) for k in self._get_keys()]
+
```
michaella21: not pretty, but it guarantees that we have a valid file path and `meatada.keyspace` inside. better idea is always welcomed

src/vivarium_public_health/dataset_manager/hdf.py
```diff
@@ -15,35 +15,24 @@
...
 
 
-def touch(path: str, append: bool):
+def touch(path: str):
     """Creates an hdf file or wipes an existing file if necessary.
 
-    If the file exists and append is not true, the existing file
-    will be destroyed. If the file exists and append is true, touch
-    will do nothing.
+    If the hdf
```
kfwilson: Looks like this comment maybe wasn't finished?

src/vivarium_public_health/dataset_manager/artifact.py
```diff
@@ -31,11 +37,11 @@ def __init__(self, path: str, filter_terms: List[str]=None):
             A set of terms suitable for usage with the ``where`` kwarg for ``pd.read_hdf``
         """
 
+        create_hdf_with_keyspace(path)
```
kfwilson: so if someone creates an Artifact and gives it a path that doesn't exist, this will now create an empty Artifact? do we want to be doing that here?

michaella21: that's what we concluded after we had an issue with initializing artifact using the `meatdata.keyspace`, if I understand correctly. It is related to with a comment from James last week regarding the encapsulation issue

tests/dataset_manager/test_hdf.py
```diff
@@ -77,31 +77,29 @@ def test_touch_no_file(mocker):
...
 
-    hdf.touch(path, False)
+    hdf.touch(path)
     tables_mock.open_file.assert_called_once_with(str(path), mode='w')
     tables_mock.reset_mock()
 
-    hdf.touch(path, True)
-    tables_mock.open_file.assert_not_called()
+    # should wipe out and make it again
+    hdf.touch(path)
```
kfwilson: Does this really test that it was wiped out and remade? Maybe add something to the hdf and then call touch again and make sure it’s gone

src/vivarium_public_health/dataset_manager/artifact.py
kfwilson: hmm now I'm thinking about this more. Does this mean that if I accidentally mistype the artifact path in my yaml config and then try to run a sim, I'm going to end up creating an empty Artifact and then getting an error that the first key I try to read isn't in the artifact rather than the FileNotFoundError I would previously get? Not a big deal but we might want to figure out how to catch it so the error's less confusing at some point

michaella21: yeah, I agree, catching error and giving a proper message are really important. But it's pretty difficult to tell from the beginning whether the artifact has proper data or just newly made empty one :( For that, we probably need a bit bigger change around the artifact, like including the list of components or so, since keyspace doesn't really tell you whether this artifact is enough for to run your configuration

src/vivarium_public_health/dataset_manager/artifact.py
```diff
@@ -31,11 +37,11 @@ def __init__(self, path: str, filter_terms: List[str]=None):
...
         self.path = path
         self._filter_terms = filter_terms
-
         self._cache = {}
-        self._keys = [EntityKey(k) for k in hdf.get_keys(self.path)]
+        self._keys = [EntityKey(k) for k in hdf.load(self.path, EntityKey('metadata.keyspace'), None)]
```
collijk: I'm thinking that it might be worthwhile to have a small `Keys` class that's responsible for maintaining the relationship between the hdf keyspace and the artifact keyspace.  E.g.:

```
class Keys:

    keyspace_node = EntityKey('metadata.keyspace')

    def __init__(self, artifact_path):
        self.artifact_path = artifact_path
        self._keys = [EntityKey(k) for k in hdf.load(self.path, EntityKey('metadata.keyspace'), None)]

    def append(self, new_key):
        new_key = EntityKey(new_key)
        self._keys.append(new_key)
        hdf.remove(self.artifact_path, self.keyspace_node)
        hdf.write(self.artifact_path, self.keyspace_node, self._keys)

   def remove(self, key):
        similar

    def to_list(self):
        return self._keys

    def __contains__(self, item):
        return EntityKey(item) in self._keys
        
```

This would make the syntax for dealing with the keys in the artifact class much simpler and separate the functionality for maintaining keyspace parity to a separate object.

michaella21:
hold on, I need to change the vivarium_inputs along with this change and may add some more tests

src/vivarium_public_health/dataset_manager/artifact.py
```diff
@@ -31,16 +31,22 @@ def __init__(self, path: str, filter_terms: List[str]=None):
             A set of terms suitable for usage with the ``where`` kwarg for ``pd.read_hdf``
         """
 
+        Artifact.create_hdf_with_keyspace(path)
```
collijk: This can (and should be) a `self.create_hdf_with_keyspace(path)`.  Pretty much always.  Even for class methods.  It doesn't matter here at all, but that's how you ensure inheritance works properly.

PR #52, opened 2018-12-27
Fix paf calculation

collijk:


src/vivarium_public_health/risks/effect.py
```diff
@@ -76,29 +76,27 @@ def adjust_target(self, index, target):
...
         if self._config_data:
             exposure = build_exp_data_from_config(builder, self.risk)
-
             rr = build_rr_data_from_config(builder, self.risk, self.affected_entity, self.affected_measure)
-            rr[self.affected_entity_type] = self.affected_entity
-
             paf_data = get_paf_data(exposure, rr)
+            paf_data[self.affected_entity_type] = self.affected_entity
         else:
             filter_name, filter_term = self.affected_entity_type, self.affected_entity
```
halfhorst: I find renaming affected_entity_type and affected_entity to filter_* down only one branch a little confusing in terms of quick readability, but I don't see a bug there

PR #56, opened 2019-01-14
Add SIR with duration model

collijk:


src/vivarium_public_health/disease/models.py
```diff
@@ -125,6 +124,49 @@ def setup(self, builder):
...
+    def __init__(self, cause: str, duration):
+        """
+        Parameters
+        ----------
+        cause
+        duration
+        """
+        self.cause = cause
+        if not isinstance(duration, pd.Timedelta):
+            self.duration = pd.Timedelta(days=float(duration) // 1, hours=(float(duration) % 1) * 24.0)
```
aflaxman: I think we are going to regret this complexity!
```suggestion
            self.duration = pd.Timedelta(days=duration)
```

src/vivarium_public_health/disease/models.py
```diff
@@ -125,6 +124,49 @@ def setup(self, builder):
...
+        else:
+            self.duration = duration
+
+    def setup(self, builder):
+        only_morbid = builder.data.load(f'cause.{self.cause}.restrictions')['yld_only']
+
+        healthy = SusceptibleState(self.cause)
+
+        get_data_functions = {}
```
kfwilson: can we name this something like disease_model_data_functions? I know this is a pattern we've had before but I found it a little confusing to see this and then have get_data_functions args to the disease state that are different. It took me a minute to figure out that this was separate from those dictionaries

halfhorst: we may as well just omit the variable and define in the dict directly like we do with dwell time

PR #57, opened 2019-01-14
Update population component for gbd2017

collijk:


src/vivarium_public_health/population/base_population.py
```diff
@@ -23,27 +18,27 @@ class BasePopulation:
     }
 
     def setup(self, builder):
-        """
-        Parameters
-        ----------
-        builder : vivarium.framework.engine.Builder
-        """
+        self.config = builder.configuration.population
+        input_config = builder.configuration.input_data
```
kfwilson: It looks like the only place this is used is to get the location on line 35 - probably don't need a separate variable for it

PR #58, opened 2019-01-14
Update crude birth rate for gbd2017

collijk:


src/vivarium_public_health/population/add_new_birth_cohorts.py
```diff
@@ -125,14 +110,48 @@ def add_new_birth_cohort(self, event):
...
+    @staticmethod
+    def get_crude_birth_rate(builder):
+        population_data = load_population_structure(builder)
+        exit_age = builder.configuration.population.to_dict().get('exit_age', None)
+        if exit_age:
+            if builder.configuration.population.age_end != exit_age:
+                raise ValueError('If you specify an exit age, the initial population age end must be the same '
+                                 'for the crude birth rate calculation to work.')
+            cut_bin = population_data[(population_data.age_group_start < exit_age)
+                                      & (population_data.age_group_end >= exit_age)]
```
kfwilson: I think you want this to be just > not >= or you want line 125 to be < and not <=. If you end up in the situation where you have a bin with age_group_end == exit_age, I think this will include it twice - once here and once in line 125

src/vivarium_public_health/population/add_new_birth_cohorts.py
```diff
@@ -125,14 +110,48 @@ def add_new_birth_cohort(self, event):
...
+        exceeds_data = builder.configuration.time.end.year > birth_rate.index.max()
+        if exceeds_data:
+            if builder.configuration.interpolation.extrapolate:
+                new_index = pd.RangeIndex(birth_rate.index.min(), builder.configuration.time.end.year + 1)
+                birth_rate = birth_rate.reindex(new_index, fill_value=birth_rate.at[birth_rate.index.max()])
+            else:
+                raise ValueError('Trying to extrapolate beyond the end of available birth data.')
+        return birth_rate
+
+    def birth_rate_producer(self, year):
```
kfwilson: Is this function used anywhere?

src/vivarium_public_health/population/add_new_birth_cohorts.py
```diff
@@ -238,4 +254,4 @@ def step(self, event):
                                             'age_end': 0,
                                         })
             parents = pd.Series(data=had_children.index, index=idx, name='parent_id')
-            self.population_view.update(parents)
+            self.population_view.update(parents)\
```
kfwilson: why is there a slash here?

PR #62, opened 2019-01-16
Feature/risk exposure observer

michaella21:
Observer component to record the exposure proportion by age-group, by year at the midpoint of the year by default.


src/vivarium_public_health/risks/observer.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(self, risk: str):
+        """
+        Parameters
+        ----------
+        risk :
+        the type and name of a risk, specified as "type.name". Type is singular.
+
+        """
+        self.risk_type, self.risk_name = split_risk_from_type(risk)
+        self.configuration_defaults = CategoricalRiskObserver.configuration_defaults
```
kfwilson: I don't think you want to just set config defaults like this because that means that you will use the same block for all risks. It seems like we want to allow for users to e.g., set different sample dates for different risks. Look at how we set config defaults in risk or risk effect to set it to the specific risk name

src/vivarium_public_health/risks/observer.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def setup(self, builder):
+        self.clock = builder.time.clock()
+        self.population_view = builder.population.get_view(['alive', 'age'])
+        self.age_bins = builder.data.load('population.age_bins')
```
kfwilson: good idea to use age bins to get the age group name - have we updated the mortality observer to do the same?

src/vivarium_public_health/risks/observer.py
```diff
@@ -0,0 +1,80 @@
...
+
+        if builder.configuration.population.exit_age:
+            self.age_bins = self.age_bins[self.age_bins.age_group_end <= builder.configuration.population.exit_age]
```
kfwilson: make sure to adjust down the last age bin if it straddles the exit age and use < not <= here - see the mortality observer

PR #67, opened 2019-02-09
update ensemble simulation distributions to handle new weights data

michaella21:


src/vivarium_public_health/risks/data_transformations.py
```diff
@@ -154,7 +154,6 @@ def get_exposure_distribution_weights(builder, risk: RiskString):
     distribution_type = get_distribution_type(builder, risk)
     if distribution_type == 'ensemble':
         weights = builder.data.load(f'{risk}.exposure_distribution_weights')
```
kfwilson: I would pivot weights here instead of in the EnsembleSimulation

src/vivarium_public_health/risks/distributions.py
```diff
@@ -4,21 +4,26 @@
...
+from vivarium_public_health.risks.data_transformations import pivot_categorical
 
 class MissingDataError(Exception):
     pass
 
 
-class EnsembleSimulation(risk_distributions.EnsembleDistribution):
+class EnsembleSimulation:
+    def __init__(self, weights, mean, sd):
+        self._weights = pivot_categorical(weights)
```
kfwilson: seems odd to pivot here - I'd do it in data transforms

src/vivarium_public_health/risks/distributions.py
```diff
@@ -4,21 +4,26 @@
...
     def setup(self, builder):
-        builder.components.add_components(self._distributions.values())
+        self.weights = builder.lookup.build_table(self._weights)
+        self.mean = builder.lookup.build_table(self._mean.drop('parameter', axis=1))
+        self.sd = builder.lookup.build_table(self._sd)
 
-    def get_distribution_map(self):
-        dist_map = super().get_distribution_map()
-        return {dist_name: partial(SimulationDistribution, distribution=dist) for dist_name, dist in dist_map.items()}
+    def ppf(self, x):
```
kfwilson: oh this is why it's so slow. You're recalculating all the distribution params on each call to this whereas before it happened once. I think you want to do it like we do SimulationDistribution where we pre-calc the params and then put those in a lookup table

src/vivarium_public_health/risks/distributions.py
```diff
@@ -94,12 +107,11 @@ def get_distribution(risk, distribution_type, exposure, exposure_standard_deviat
         distribution = PolytomousDistribution(risk, exposure)
     elif distribution_type == 'normal':
         distribution = SimulationDistribution(mean=exposure, sd=exposure_standard_deviation,
-                                              distribution=risk_distributions.Normal)
+                                              distribution=Normal)
```
kfwilson: do we need to wrap these too so their ppf's fill in nulls with 0s?

PR #73, opened 2019-02-13
no val column

kfwilson:
this was breaking things for eggs because we pivot categorical risk data so there's no longer a value column. I tested by running sanofi to make sure that this won't break things for cont risks


halfhorst:
So, this broke obesity

PR #74, opened 2019-02-13
drop parameter column

halfhorst:
parameter column needs to not make its way in to the lookup table for continuous risks


kfwilson:
is the param column in exposure data fine? that's the other place where we're pivoting only categorical risks so we'll end up wide on parameter only for categorical but with a value and a param column for continuous risks

PR #77, opened 2019-02-14
Feature/mort by sex

halfhorst:
I did this the fast way, not the way I'd like. This splits by sex by default. I would like for this observer (and my  BMI ones) to have switches for year, age and sex that you can turn on and off but that will take a bit of work because of the way the work is split across functions here. I just need by sex to work for now. Also, perhaps the observer should only report back age groups that cover ages in the sim (meaning excluding younger ages if the start age is, say 20).

I split get_sexes into a utility function as well. I don't anticpate sexes changing but I wanted them somewhere where we could change them if needed, or pull them from inputs or wherever

### DIFF
--- a/src/vivarium_public_health/metrics/mortality.py
+++ b/src/vivarium_public_health/metrics/mortality.py
@@ -1,7 +1,7 @@
 import pandas as pd
 import numpy as np
 
-from .utilities import get_age_bins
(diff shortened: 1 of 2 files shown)


halfhorst:
On second thought, I could just split by sex like this in a bmi-specific observer and then do this right after.

PR #78, opened 2019-02-14
Feature/treatment observer

michaella21:
treatment observer, counts each dose of vaccine by year

### DIFF
--- a/src/vivarium_public_health/metrics/treatment.py
+++ b/src/vivarium_public_health/metrics/treatment.py
@@ -17,6 +17,16 @@ class TreatmentObserver:
     and years by setting the ``by_age``, ``by_sex``, and ``by_year``
     flags, respectively.
 
+    In the model specification, your configuration for this component should
+    be specified as, e.g.:
+
+    configuration:
+        metrics:
+            YOUR_TREATMENT_NAME:
+                by_age: True
+                by_year: False
+                by_sex: True
+
     """
 
     configuration_defaults = {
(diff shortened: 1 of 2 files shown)


collijk:
James to handle merge into old and new branches

src/vivarium_public_health/metrics/treatment.py
```diff
@@ -0,0 +1,73 @@
...
+    By default, this observer computes aggregate dose counts over the entire
+    simulation.  It can be configured to bin these into age_groups, sexes,
+    and years by setting the ``by_age``, ``by_sex``, and ``by_year``
+    flags, respectively.
+
+    """
+
+    configuration_defaults = {
+        'metrics': {
+            'treatment_observer': {
```
halfhorst: I could see myself uses "treatment_observer" directly in the config and getting default behavior accidently. May want to add a note that the config block should be specified with the passed treatment name

michaella21: I agree. Since I'm also confused by this, I tend to add the example to show how config should be given in docstring, not actually for the user, really for reminding myself.

michaella21:
and.. again I can't approve this according to the github :(

collijk:
haha, sorry michelle!

halfhorst:
> and.. again I can't approve this according to the github :(

Maybe it doesn't let you approve your own PRs

collijk:
That's totally the issue

PR #81, opened 2019-02-20
Feature/data free rr dist

kfwilson:


src/vivarium_public_health/risks/data_transformations.py
```diff
@@ -208,29 +210,41 @@ def load_relative_risk_data(builder, risk: RiskString, target: TargetString):
     distribution_type = get_distribution_type(builder, risk)
     relative_risk_source = builder.configuration[f'effect_of_{risk.name}_on_{target.name}'][target.measure]
 
-    if relative_risk_source == 'data':
+    if relative_risk_source['relative_risk'] is not None:
+        relative_risk_data = _make_relative_risk_data(builder, float(relative_risk_source['relative_risk']))
+    elif relative_risk_source['log_mean'] is not None:  # log distribution
```
halfhorst: should we do some checking to make sure someone doesn't specify a mean and a log_mean and get a possibly unexpected result?

src/vivarium_public_health/risks/data_transformations.py
```diff
@@ -198,39 +200,54 @@ def rebin_exposure_data(builder, risk: RiskString, data: pd.DataFrame):
...
     return relative_risk_data
 
 
-def load_relative_risk_data(builder, risk: RiskString, target: TargetString):
+def load_relative_risk_data(builder, risk: RiskString, target: TargetString, source_type: str):
     distribution_type = get_distribution_type(builder, risk)
     relative_risk_source = builder.configuration[f'effect_of_{risk.name}_on_{target.name}'][target.measure]
 
-    if relative_risk_source == 'data':
+    if source_type == 'relative risk value':
```
kfwilson: @collijk I left this as the branching structure rather than pulling out functions and making a dict mapping source type to function because I needed different parameters depending on source type so it ended up being a little messy with a lot of partial function calls. I can definitely go to that if you'd rather keep the dict pattern but I thought this ended up being more readable

PR #83, opened 2019-02-21
Feature/cbr fertility update

collijk:


src/vivarium_public_health/population/add_new_birth_cohorts.py
```diff
@@ -56,50 +49,52 @@ def add_new_birth_cohort(self, event):
...
     textbooks.
     .. _website: https://stats.oecd.org/glossary/detail.asp?ID=490
     .. _Wikipedia: https://en.wikipedia.org/wiki/Birth_rate
     """
+
+    configuration_defaults = {
+        'fertility': {
+            'time_dependent_live_births': True,
+            'time_dependent_population_fraction': False,
+        }
```
michaella21: It definitely needs some comments. I don't really get what these flags mean unless I dig down the codes.

src/vivarium_public_health/population/add_new_birth_cohorts.py
```diff
@@ -56,50 +49,52 @@ def add_new_birth_cohort(self, event):
...
-            A function or method for creating a population.
-        Notes
-        -----
-        The method for computing the Crude Birth Rate employed here is
-        approximate.
         """
-        birth_rate = self.birth_rate.at[event.time.year]
-        population_size = len(event.index)
-        step_size = event.step_size / pd.Timedelta(seconds=1)
+        birth_rate = self.birth_rate.at[self.clock().year]
```
michaella21: was it the change that we decided to do yesterday during sprint planning, right? just checking whether I'm not missing.

collijk: Yeah, that's right.

PR #84, opened 2019-02-21
Update old code with new fertility component

collijk:
This looks bigger than it is since it copies some other updates already on develop.

### DIFF
--- a/src/vivarium_public_health/population/add_new_birth_cohorts.py
+++ b/src/vivarium_public_health/population/add_new_birth_cohorts.py
@@ -49,6 +49,19 @@ class FertilityDeterministic:
 
 class FertilityCrudeBirthRate:
     """Population-level model of births using crude birth rate.
+    The number of births added each time step is calculated as
+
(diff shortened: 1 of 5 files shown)


src/vivarium_public_health/population/add_new_birth_cohorts.py
```diff
@@ -168,17 +109,13 @@ class FertilityAgeSpecificRates:
...
         ----------
         builder : vivarium.engine.Builder
             Framework coordination object.
-
         """
-
         self.randomness = builder.randomness.get_stream('fertility')
-        asfr_data = builder.data.load("covariate.age_specific_fertility_rate.estimate",
-                                      future=builder.configuration.input_data.forecast)
+        asfr_data = builder.data.load("covariate.age_specific_fertility_rate.estimate")
```
kfwilson: if this is old code, I think we want to keep the future flag or building artifacts for diarrhea won't use the forecast asfr data

collijk: Good catch

PR #85, opened 2019-02-21
Add mortality observer to old code

collijk:


src/vivarium_public_health/metrics/mortality.py
```diff
@@ -0,0 +1,137 @@
...
+
+def count_person_time(pop, age_bins, start_time, end_time):
+    lived_in_span = pop[(start_time <= pop.exit_time) & (pop.entrance_time < end_time)]
+    # The right way to do this is np.maximum/np.minimum,
+    # but there's some bug in pandas that causes that to break.
+    entrance_time = lived_in_span.entrance_time
+    exit_time = lived_in_span.exit_time
+    exit_time.loc[end_time < exit_time] = end_time
+
+    years_in_span = (exit_time - entrance_time) / pd.Timedelta(days=365.25)
```
kfwilson: why not use your to_years function here?

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -0,0 +1,137 @@
...
+    lived_in_span['age_at_start'] = np.maximum(lived_in_span.age - years_in_span, 0)
+
+    data = pd.Series(0, index=age_bins.index)
+
+    for group, age_bin in age_bins.iterrows():
+        start, end = age_bin.age_group_start, age_bin.age_group_end
+        in_group = lived_in_span[(start < lived_in_span.age) & (lived_in_span.age_at_start < end)]
```
kfwilson: at some point, some in line comments might be really helpful to add. It took me a bit to figure out why this was < instead of <=

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class MortalityObserver:
+    """ An observer for total and cause specific deaths during simulation.
+    This component by default observes the total number of deaths out of
+    total population and total person time that each simulant spent
+    during the simulation until it exits.
+    These data are categorized by age groups and causes and aggregated over
+    total population as well as the population who were born during
+    the simulation.
```
kfwilson: I find this doc string kind of confusing. From the code, it seems like deaths and person time are always grouped by age (i.e., not aggregated over total population) but regardless of whether or not you have the by_year flag turned on, you will get deaths and person time aggregated over the whole course of the sim. If you have the by_year flag turned on, you will additionally get deaths and person time by year. Reading this doc string, I am confused about what results I'm going to get by what group. 

also: why do we only group by age/year and not sex? is that not of interest?

halfhorst:
this is pretty much a direct copy, right ?

PR #91, opened 2019-02-25
Feature/birth prevalence

kfwilson:
This PR includes the necessary changes to support birth prevalence. One thing to note is that this changes the previous behavior for non-birth prevalence models where previously newly born simulants were initialized into disease states based on prevalence data but now those simulants are initialized based on default birth prevalence data of 0. 

This includes 2 neonatal models: susceptible/with condition with and without incidence. I'm waiting to hear from Christine about whether she will also need models with duration/remission but I can do those in a separate PR.


src/vivarium_public_health/disease/model.py
```diff
@@ -109,7 +112,21 @@ def load_population_columns(self, pop_data):
 
         assert self.initial_state in {s.state_id for s in self.states}
 
-        state_names, weights_bins = self.get_state_weights(pop_data.index)
+        # FIXME: this is a hack to figure out whether or not we're at the simulation start based on the fact that the
```
collijk: Make a ticket for this fixme.  I think what we should do is always provide the user_data with a key indicating something about the state of the system when people are being created.  It should be a relatively small fix (one change in the engine, and the rest in the fertility component).

halfhorst:
>  but now those simulants are initialized based on default birth prevalence data of 0.

This means no one is initialized with condition, right ?

src/vivarium_public_health/disease/model.py
kfwilson: ticket created

kfwilson:
> > but now those simulants are initialized based on default birth prevalence data of 0.
> 
> This means no one is initialized with condition, right ?

Yes. Previously those sims were initialized based on prevalence data (which may or may not have had 0 for age 0) but now they will always be initialized to susceptible.

halfhorst:
One other thing -- the allclose fails pretty frequently for me. Is it consistently passing for you ?

PR #94, opened 2019-02-27
bugfix in get_live_births_per_year

michaella21:
2016 covariates and forecasting covariates don't have a parameter or value columns thus it broke here.


michaella21:
yep, I tested all 4 combinations with two flags on and off before I opened this PR.

PR #96, opened 2019-02-28
Feature/rebin risks

kfwilson:
Here's configurable rebinning for risks. Added unit tests for the rebin rr and exp functions + tested the whole thing with bfp

### DIFF
--- a/src/vivarium_public_health/risks/base_risk.py
+++ b/src/vivarium_public_health/risks/base_risk.py
@@ -21,12 +21,28 @@ class Risk:
     (2) configuration:
             risk:
                 exposure: proxy_covariate
+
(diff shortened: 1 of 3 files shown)


src/vivarium_public_health/risks/base_risk.py
```diff
@@ -21,12 +21,28 @@ class Risk:
...
     """
 
     configuration_defaults = {
         "risk": {
             "exposure": 'data',
-            "rebin": {},
+            "rebin": {
+                "exposed": [],
+            },
             "category_thresholds": [],
```
kfwilson: what's this for? I was going to update the docstring to mention it when I updated it for rebin but I couldn't remember what this was for

PR #97, opened 2019-02-28
Feature/rebin risks

kfwilson:
implements configurable risk rebinning (user supplies list of categories to be rebinned into a single exposed category via the config). Tested with bfp + added unit tests for individual exp/rr rebin functions


src/vivarium_public_health/risks/base_risk.py
```diff
@@ -22,12 +22,28 @@ class Risk:
...
     """
 
     configuration_defaults = {
         "risk": {
             "exposure": 'data',
-            "rebin": {},
+            "rebin": {
+                "exposed": [],
+            },
             "category_thresholds": [],
```
kfwilson: what is category_thresholds for? I was going to add it to the docstring when I updated it for rebin but I couldn't remember what this was

src/vivarium_public_health/risks/data_transformations.py
```diff
@@ -7,11 +7,12 @@
 from vivarium_public_health.risks import distributions
 from vivarium_public_health.utilities import EntityString, TargetString
 
+REBIN_UNSUPPORTED = ['unsafe_water_source', 'low_birth_weight_and_short_gestation']
```
collijk: Why are these unsupported?

src/vivarium_public_health/risks/base_risk.py
collijk: They are thresholds we use to bin the continuous distributions for alternative risks.

collijk: They should be mutually exclusive with the rebin arguments.

src/vivarium_public_health/risks/base_risk.py
```diff
@@ -22,12 +22,28 @@ class Risk:
...
+        risk:
+            rebin:
+                exposed: ['cat1', 'cat2']
     """
 
     configuration_defaults = {
         "risk": {
             "exposure": 'data',
-            "rebin": {},
+            "rebin": {
```
collijk: Do we need to nest the argument here?  Maybe just a top level key of  ``rebinned_exposed``?

src/vivarium_public_health/risks/data_transformations.py
kfwilson: this was a hold over from the old code. I wasn't sure if there was some conceptual problem with rebinning them? I can try them and see what happens

src/vivarium_public_health/risks/base_risk.py
kfwilson: nope I just wasn't thinking. changed to top level key

src/vivarium_public_health/risks/base_risk.py
kfwilson: ok I added a bit to the docstring about this. will you make sure I represented it correctly? And I added a check to the rebinned validator to make sure that category thresholds aren't also supplied

halfhorst:
Barring the potentially unnecessary exceptions it looks good

src/vivarium_public_health/risks/data_transformations.py
michaella21: Isn't it the two 2-D risk factors? and I don't think that we have a good idea how to rebin this unless researchers figure out how.

kfwilson: so I'm trying with these risk factors now to see if it works (it's taking a while because I don't have caches built for these), but even if I can, should I leave these as unsupported with a note about their being 2D risk factors?

collijk: Since you made it configurable, they could do this if they wanted.

src/vivarium_public_health/risks/base_risk.py
```diff
@@ -22,12 +22,35 @@ class Risk:
...
+    dichotomous risk. For example, for a risk named "risk" with categories
+    cat1, cat2, cat3, and cat4 that you wished to rebin into a dichotomous risk
+    with an exposed category containing cat1 and cat2 and an unexposed category
+    containing cat3 and cat4, the configuration could look like this:
+
+    configuration:
+        risk:
+            rebinned_exposed: ['cat1', 'cat2']
+
+    For alternative risk factors, you can provide a 'category_thresholds'
```
collijk: "can" should be "must"

src/vivarium_public_health/risks/data_transformations.py
collijk: Since you made these configurable, everything should be fine.  The users can rebin risks however they want.

collijk: Oh, I forgot.  The thing that we want for rebinned risks is to take multiple categories to multiple categories.  e.g.
`rebinned_exposure = {'cat1': ['cat1', 'cat2'], 'cat2': ['cat3'], 'cat3': ['cat4', 'cat5']}`.  This would allow someone to, for instance, collapse one of the dimensions of unsafe water, or to take lbwsg and turn it into a 4-category risk instead of a 50 category one.

This is outside the scope of this ticket, but make sure it's on a ticket somewhere.

src/vivarium_public_health/risks/base_risk.py
kfwilson: done

PR #98, opened 2019-03-01
[WIP] First draft of mortality observer update

collijk:
Still working on cleaning this up and separating into single purpose functions.

### DIFF
--- a/src/vivarium_public_health/metrics/disease.py
+++ b/src/vivarium_public_health/metrics/disease.py
@@ -1,6 +1,7 @@
 from collections import defaultdict
+from string import Template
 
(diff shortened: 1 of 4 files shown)


michaella21:
While you're now working on all these observers, could you make sure that all the metric column name has only lowercase instead of mix of lower and upper cases? According to Derrick, Mortality observer returns mixed cases like `person_time_in_2010_among_Early_Neonatal_born_in_sim` and Risk observer returns all lower cases like `child_underweight_cat1_exposed_in_2010_among_early_neonatal`, which confused people when they had so many columns in the output.

PR #99, opened 2019-03-01
Put in query string

collijk:
Splitting up the mortality observer changes and writing tests.


halfhorst:
this seems like it will be more broadly useful

PR #103, opened 2019-03-04
Pull out calculation functions from disease observer

collijk:
Tests on the way

### DIFF
--- a/src/vivarium_public_health/metrics/utilities.py
+++ b/src/vivarium_public_health/metrics/utilities.py
@@ -144,7 +144,7 @@ def get_group_counts(pop: pd.DataFrame, base_filter: str, base_key: Template,
             template_kwargs = {'sex': sex.lower(), 'age_group': group.lower()}
             key = Template(base_key.safe_substitute(**template_kwargs))
             group_filter = base_filter.format(**filter_kwargs)
-            in_group = pop.query(group_filter)
+            in_group = pop.query(group_filter) if group_filter else pop
 
             group_counts[key] = len(in_group)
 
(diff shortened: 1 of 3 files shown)


tests/metrics/test_utilities.py
```diff
@@ -46,3 +78,66 @@ def test_query_string(a, b):
...
+    expected_value = len(pop) / 2
+    if observer_config['by_sex']:
+        expected_value /= 2
+    if observer_config['by_age']:
+        expected_value /= len(age_bins)
+
+    # Doubling pop should double person time
+    pop = pd.concat([pop, pop], axis=0, ignore_index=True)
+
+    person_time = get_disease_event_counts(pop, observer_config, disease, event_time, age_bins)
```
kfwilson: this should probably be counts instead of person_time?

tests/metrics/test_utilities.py
```diff
@@ -1,8 +1,40 @@
...
+
+
+@pytest.fixture(params=((0, 100, 5, 1000), (20, 100, 5, 1000)))
+def ages_and_bins(request):
+    age_min = request.param[0]
+    age_max = request.param[1]
+    age_groups = request.param[2]
+    num_ages = request.param[3]
+
+    ages = np.linspace(age_min, age_max - age_groups/num_ages, num_ages)
```
kfwilson: ooh I like this. So much better than manually making up data (my bad goto!)

src/vivarium_public_health/metrics/utilities.py
```diff
@@ -80,18 +80,18 @@ def get_output_template(by_age: bool, by_sex: bool, by_year: bool) -> Template:
         A template string with measure and possibly additional criteria.
 
     """
-    template = '{measure}'
+    template = '${measure}'
```
kfwilson: I haven't really used template strings before so why do you need the $?

collijk: Holdover from old string formatting, I think.  I was surprised too.  It's just the escape character used here.  I think we could customize that, but not worth it.

tests/metrics/test_utilities.py
collijk: Building data is always hard.

tests/metrics/test_utilities.py
collijk: Will fix in the next PR

PR #105, opened 2019-03-05
constrain propensity to valid range

halfhorst:
This turned up in the obesity sim -- we noticed BMIs of zero. Turns out we sample propensities that fall outside of the range we consider valid for returning values from .ppf(). Those propensities result in returned values of NaN that then get set to zero, which doesn't make sense. 

My only issue is that those constants feel magical.

### DIFF
--- a/src/vivarium_public_health/risks/distributions.py
+++ b/src/vivarium_public_health/risks/distributions.py
@@ -29,9 +29,9 @@ class EnsembleSimulation:
     def ppf(self, q):
         if not q.empty:
             # We limit valid propensity to [0.001 0.999]. Beyond that bound values return NaN and then become zero,
-            # which is nonsensical.
-            q[q > 0.999] = 0.998
-            q[q < 0.001] = 0.0011
+            # which is nonsensical. We avoid the inclusive limit to protect ourselves from a math error.
+            q[q >= 0.999] = 0.998
+            q[q <= 0.001] = 0.0011
             weights = self.weights(q.index)
             parameters = {name: parameter(q.index) for name, parameter in self.parameters.items()}
             x = EnsembleDistribution(weights, parameters).ppf(q)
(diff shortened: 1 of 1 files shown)


src/vivarium_public_health/risks/distributions.py
```diff
@@ -28,6 +28,10 @@ def _get_parameters(self, weights, mean, sd):
 
     def ppf(self, q):
         if not q.empty:
+            # We limit valid propensity to (0.001 0.999). Beyond that bound values return NaN and then become zero,
+            # which is nonsensical.
+            q[q > 0.999] = 0.998
```
kfwilson: can propensity be 0.999? Your comment makes it seem like the range is exclusive, but then only bump down outside of the end points (making it seem inclusive)?

halfhorst: I'm unsure, because I can't actually see where the limit spawns from, but I bumped inside the limit to be safe, and I made the comment show exclusive to be safe as well

halfhorst: I'll test it

halfhorst: It's appears to be inclusive. I think the limit comes from a percentage of a lognormal distribution. I think not brushing against the limit is safer, though, and is what james and I discussed

kfwilson: Ok I guess I'm just confused over whether or not a propensity of 0.999 (which won't be adjusted down to 0.998 but from the comment seems to be invalid) will result in the NaN -> 0 issue

kfwilson: Oh nvm! Just saw the updated comment.

halfhorst: yeah that's fair. it's weird to leave 999 and resassign greater than 999 to 998. I'll just make it inclusive.

PR #106, opened 2019-03-06
Feature/reset indices

halfhorst:
This resets indices in the last controller, the artifact manager.There is also a change to the function that writes to the HDF. We no longer write data_columns, we have indices now. 

I tested this by running simulate on the obesity yaml. The HDF write was tested in the vivarium_inputs PR by building artifacts for the smoke test configurations

EDIT:  We now support writing empty dataframes and loading them back to the same state. Re-tested on obesity.

### DIFF
--- a/src/vivarium_public_health/dataset_manager/dataset_manager.py
+++ b/src/vivarium_public_health/dataset_manager/dataset_manager.py
@@ -111,6 +111,7 @@ class ArtifactManager:
             if the data is a dataframe.
         """
         data = self.artifact.load(entity_key)
+        # demog dimensions is a df but not multi-indexed because it has no columns beyond our indexing dimensions
         if isinstance(data, pd.DataFrame) and isinstance(data.index, pd.MultiIndex):
             data = data.reset_index()
         return filter_data(data, self.config_filter_term, **column_filters) if isinstance(data, pd.DataFrame) else data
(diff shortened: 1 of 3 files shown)


src/vivarium_public_health/dataset_manager/dataset_manager.py
```diff
@@ -111,6 +111,8 @@ def load(self, entity_key: str, **column_filters: _Filter):
             if the data is a dataframe.
         """
         data = self.artifact.load(entity_key)
+        if isinstance(data, pd.DataFrame) and isinstance(data.index, pd.MultiIndex):
```
halfhorst: will add same comment here as in vivarium_inputs

src/vivarium_public_health/dataset_manager/hdf.py
```diff
@@ -139,13 +137,10 @@ def _write_data_frame(path: str, entity_key: 'EntityKey', data: pd.DataFrame):
     if data.empty:
         raise ValueError("Cannot persist empty dataset")
 
-    # Even though these get called data_columns, it's more correct to think of them
-    # as the columns you can use to index into the raw data with. It's the subset of columns
-    # that you can filter by without reading in a whole dataset.
-    data_columns = DEFAULT_COLUMNS.intersection(data.columns)
-
+    # Our data is indexed (mostly -- demog dimensions isn't).
```
kfwilson: hmm so here's something I just thought of: we will now no longer be able to filter demog dimensions by location without reading it all in. Is that going to cause problems for multi-loc artifacts?

PR #108, opened 2019-03-08
Age, sex, and time filters

collijk:
These will be used in the mortality observer as well.  Bringing them in now to keep prs manageable.

### DIFF
--- a/src/vivarium_public_health/metrics/utilities.py
+++ b/src/vivarium_public_health/metrics/utilities.py
@@ -262,9 +262,8 @@ def get_group_counts(pop: pd.DataFrame, base_filter: str, base_key: OutputTempla
     for group, age_group in ages:
         start, end = age_group.age_group_start, age_group.age_group_end
         for sex in sexes:
-            filter_kwargs = {'age_group_start': start, 'age_group_end': end, 'sex': sex}
-            template_kwargs = {'sex': sex.lower(), 'age_group': group.lower()}
-            key = base_key.substitute(**template_kwargs)
+            filter_kwargs = {'age_group_start': start, 'age_group_end': end, 'sex': sex, 'age_group': group}
+            key = base_key.substitute(**filter_kwargs)
             group_filter = base_filter.format(**filter_kwargs)
             in_group = pop.query(group_filter) if group_filter else pop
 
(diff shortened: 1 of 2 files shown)


src/vivarium_public_health/metrics/utilities.py
```diff
@@ -148,6 +149,85 @@ def get_output_template(by_age: bool, by_sex: bool, by_year: bool) -> OutputTemp
...
+        The time the simulation starts.
+    sim_end
+        The time the simulation ends.
+
+    Returns
+    -------
+    span_filter
+        A filter on time for use with DataFrame.query
+    time_spans
+        Iterable for the age and sex groups partially defining the bins
```
kfwilson: should this be year groups instead of age and sex?

src/vivarium_public_health/metrics/utilities.py
```diff
@@ -148,6 +149,85 @@ def get_output_template(by_age: bool, by_sex: bool, by_year: bool) -> OutputTemp
...
+        for the observers.
+
+    """
+    if config['by_year']:
+        time_spans = [(year, (pd.Timestamp(year=year, month=1, day=1), pd.Timestamp(f'1-1-{year + 1}')))
```
kfwilson: Is there a reason the year start timestamp is constructed differently than all the others?

src/vivarium_public_health/metrics/utilities.py
```diff
@@ -148,6 +149,85 @@ def get_output_template(by_age: bool, by_sex: bool, by_year: bool) -> OutputTemp
...
+                      for year in range(sim_start.year, sim_end.year + 1)]
+    else:
+        time_spans = [('all_years', (pd.Timestamp(f'1-1-1900'), pd.Timestamp(f'1-1-2100')))]
+    # This filter needs to be applied separately to compute additional
+    # attributes in the person time calculation.
+    span_filter = QueryString('{t_start} <= exit_time and entrance_time < {t_end}')
```
kfwilson: I think I'm still a little fuzzy on how the year binning works. Why is this <= exit_time instead of < exit_time? do we assume exit_time is the end of the day or something?

src/vivarium_public_health/metrics/utilities.py
halfhorst: I prefer this verbose one. The other one, is it american-style or everyone else- style ?

src/vivarium_public_health/metrics/utilities.py
collijk: yes

src/vivarium_public_health/metrics/utilities.py
collijk: Oh, I was trying to debug an issue and pushed this accidentally.

PR #109, opened 2019-03-08
Add lived in span and tests

collijk:
Here's a visual representation of all the unique cases: 

IMG_20190308_154720

### DIFF
--- a/src/vivarium_public_health/metrics/utilities.py
+++ b/src/vivarium_public_health/metrics/utilities.py
@@ -348,8 +348,10 @@ def get_lived_in_span(pop: pd.DataFrame, t_start: pd.Timestamp, t_end: pd.Timest
     span_exit_time = lived_in_span.exit_time.copy()
     span_exit_time.loc[t_end < span_exit_time] = t_end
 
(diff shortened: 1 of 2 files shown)


src/vivarium_public_health/metrics/utilities.py
```diff
@@ -318,6 +318,43 @@ def get_treatment_counts(pop, config, treatment, doses, event_time, age_bins):
...
+        The date and time at the end of the span.
+
+    Returns
+    -------
+    lived_in_span
+        A table representing the population who lived some amount of time
+        within the time span with all columns provided in the original
+        table and additional columns 'age_at_start' and 'age_at_end' indicating
+        the age of the individual at the start and end of the time span,
+        respectively.
```
halfhorst: if they die in the span I can assume age at end is set to death age ?

collijk: This function really doesn't know anything about why a simulant is no longer being tracked by the simulation, just that they aren't being tracked.  Should work for people who migrate in/out, exit because they get to old, etc.

halfhorst: I shouldn't have used born/died. This was just a language thing but it's good as is, no one probably interprets it like I do :stuck_out_tongue_closed_eyes:

PR #110, opened 2019-03-11
Make age filter time span aware

collijk:
For the person time calculation, we don't care about a person's age, but about whether a person has lived through an age group during a particular period of time (either the whole simulation or a subset of it).  This PR allows the age portion of the filter to handle that case as well.

### DIFF
--- a/src/vivarium_public_health/metrics/utilities.py
+++ b/src/vivarium_public_health/metrics/utilities.py
@@ -42,7 +42,10 @@ class QueryString(str):
 
 class SubstituteString(str):
     """Normal string plus a no-op substitute method.
-    Meant to be used with the OutputTemplate."""
+
+    Meant to be used with the OutputTemplate.
+
+    """
     def substitute(self, *args, **kws):
         return self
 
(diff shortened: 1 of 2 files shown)


src/vivarium_public_health/metrics/utilities.py
```diff
@@ -177,7 +184,10 @@ def get_age_sex_filter_and_iterables(config: dict, age_bins: pd.DataFrame) -> (
     age_sex_filter = QueryString("")
     if config['by_age']:
         ages = age_bins.set_index('age_group_name').iterrows()
-        age_sex_filter += '{age_group_start} <= age and age < {age_group_end}'
+        if in_span:
+            age_sex_filter += '{age_group_start} < age_at_span_end and age_at_span_start < {age_group_end}'
```
kfwilson: should this be age_group_start <= age_at_span_end since we have the start of age bins being inclusive? or is the end of the span exclusive?

halfhorst: Seconded. Unless it is exclusive I think it should count as in span based on our inclusive lower bound standard. Though our age precision means this will be rare, right ?

PR #113, opened 2019-03-12
Feature/get deaths

collijk:


src/vivarium_public_health/metrics/utilities.py
```diff
@@ -442,6 +442,55 @@ def get_person_time_in_span(lived_in_span: pd.DataFrame, base_filter: QueryStrin
...
+        represents a particular demographic subgroup.
+
+    """
+    base_filter = QueryString('alive == "dead" and cause_of_death == "{cause}"')
+    base_key = get_output_template(**config)
+    pop = clean_cause_of_death(pop)
+
+    causes = [c for c in pop.cause_of_death.unique()]
+    time_spans = get_time_iterable(config, sim_start, sim_end)
+
```
kfwilson: these are so much cleaner and more readable with all the templating!

collijk: Yeah.  There's a lot of duplicated logic though.  After I bring in all the functionality (and make a release) I'm going to spend a little time moving stuff around to try to get rid of code.

PR #118, opened 2019-03-13
Update disability observer, and fix some docstrings

collijk:


src/vivarium_public_health/metrics/disability.py
```diff
@@ -1,38 +1,92 @@
...
-def _disability_post_processor(value, step_size):
-    return rescale_post_processor(joint_value_post_processor(value, step_size), step_size)
+from .utilities import get_age_bins, get_years_lived_with_disability
 
 
 class Disability:
-    """Measures and assigns disability to the simulants."""
+    """Counts years lived with disability.
+
+    By default, this counts both aggregate and cause-specific years lived
```
kfwilson: To make sure I’m understanding this, we have both cause specific and total because total != sum of cause specific because of potential overlaps. Why do we only bin cause specific and not total ylds?

collijk: Because no one asked for them.  Let me think about whether that's easy to do or not.

PR #134, opened 2019-03-28
Rollback artifact indexing changes

collijk:


tests/dataset_manager/test_hdf.py
```diff
@@ -49,10 +48,7 @@ def hdf_file(hdf_file_path):
 _KEYS = ['population.age_bins',
          'population.structure',
          'population.theoretical_minimum_risk_life_expectancy',
-         'cause.all_causes.restrictions',
-         'metadata.versions',
```
kfwilson: I don't know if we care right now, but some of the changes that I made to the test artifact (ie.., adding these metadata keys) didn't actually have anything to do with the indices - they were changes that were required by the Artifact changes Michelle made a while ago to track versions and have the keyspace as a key in the artifact.

collijk: Nah, don't care for now.

PR #135, opened 2019-03-29
Feature/point prevalence observation

collijk:
- Cleanup some of the template usage
- Add prevalent disease counts to disease observer
- Add population count observer

### DIFF
--- a/src/vivarium_public_health/metrics/population.py
+++ b/src/vivarium_public_health/metrics/population.py
@@ -6,22 +6,26 @@ from .utilities import get_age_bins, get_population_counts
 
 
 class PopulationObserver:
-    """ An observer for cause-specific deaths, ylls, and total person time.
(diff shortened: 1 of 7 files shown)


src/vivarium_public_health/metrics/population.py
```diff
@@ -0,0 +1,79 @@
...
+                    sample_date:
+                        month: 4
+                        day: 10
+
+    """
+    configuration_defaults = {
+        'metrics': {
+            'population': {
+                'by_age': False,
+                'by_year': False,
```
kfwilson: having by_year here is a little misleading, especially because by_year: False doesn't actually do anything for this observer. If you want to keep it for consistency with the other observers, maybe add a comment indicating that this flag isn't used for this observer/is always overwritten to True in the actual observing?

PR #140, opened 2019-04-15
Feature/disease observer paf 1

halfhorst:
This updates the risk attributable disease to produce columns necessary for use with the disease observer. All it needed was the disease event time column. I added a susceptible event time column as well because recovery is afforded by the model.

I tested this on the eggs yaml below but I removed all the promotion stuff. Dependencies are a pain, we really need that manager.

https://stash.ihme.washington.edu/projects/CSTE/repos/breastfeeding_promotion_and_eggs/browse/breastfeeding_promotion_and_eggs/model_specs/eggs.yaml

### DIFF
--- a/src/vivarium_public_health/disease/special_disease.py
+++ b/src/vivarium_public_health/disease/special_disease.py
@@ -136,7 +136,9 @@ class RiskAttributableDisease:
         self.population_view.update(pop)
 
     def get_exposure_filter(self, distribution, exposure_pipeline, threshold):
+
         if distribution in ['dichotomous', 'ordered_polytomous', 'unordered_polytomous']:
+
             def categorical_filter(index):
                 exposure = exposure_pipeline(index)
                 return exposure.isin(threshold)
--- a/src/vivarium_public_health/disease/special_disease.py
+++ b/src/vivarium_public_health/disease/special_disease.py
@@ -86,6 +86,8 @@ class RiskAttributableDisease:
     def setup(self, builder):
         self.recoverable = builder.configuration[self.cause.name].recoverable
 
+        self.clock = builder.time.clock()
+
         if builder.configuration[self.cause.name].mortality:
             csmr_data = builder.data.load(f'cause.{self.cause.name}.cause_specific_mortality')
             builder.value.register_value_modifier('csmr_data', lambda: csmr_data)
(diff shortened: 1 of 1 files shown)


src/vivarium_public_health/disease/special_disease.py
```diff
@@ -121,7 +130,9 @@ def on_time_step(self, event):
         #  if this is recoverable, anyone who gets lower exposure in the event goes back in to susceptible status.
         if self.recoverable:
             pop.loc[~sick, self.cause.name] = f'susceptible_to_{self.cause.name}'
+            pop.loc[~sick, self.susceptible_event_time_column] = event.time
```
collijk: Event time columns track changes in state.  This is going to get everyone in a state and update their event time every time step, which is not what we want.

src/vivarium_public_health/disease/special_disease.py
```diff
@@ -104,23 +106,32 @@ def setup(self, builder):
...
 
     def on_initialize_simulants(self, pop_data):
-        new_pop = pd.Series(f'susceptible_to_{self.cause.name}', index=pop_data.index, name=self.cause.name)
+        new_pop = pd.DataFrame({self.cause.name: f'susceptible_to_{self.cause.name}',
+                                self.diseased_event_time_column: pd.Series(pd.NaT, index=pop_data.index),
+                                self.susceptible_event_time_column: pd.Series(pd.NaT, index=pop_data.index)},
+                               index=pop_data.index)
         sick = self.filter_by_exposure(pop_data.index)
-        new_pop[sick] = self.cause.name
+        new_pop.loc[sick, self.cause.name] = self.cause.name
```
kfwilson: do you need to set the diseased_event_time column here?

src/vivarium_public_health/disease/special_disease.py
```diff
@@ -104,24 +108,38 @@ def setup(self, builder):
...
     def on_time_step(self, event):
         pop = self.population_view.get(event.index, query='alive == "alive"')
         sick = self.filter_by_exposure(pop.index)
         #  if this is recoverable, anyone who gets lower exposure in the event goes back in to susceptible status.
         if self.recoverable:
+            change_to_susceptible = (~sick) & (pop[self.cause.name] != f'susceptible_to_{self.cause.name}')
+            pop.loc[change_to_susceptible, self.susceptible_event_time_column] = event.time
             pop.loc[~sick, self.cause.name] = f'susceptible_to_{self.cause.name}'
+        change_to_diseased = sick & (pop[self.cause.name] != self.cause.name)
+        pop.loc[change_to_diseased, self.diseased_event_time_column] = event.time
```
kfwilson: it shouldn't matter in terms of what the state table actually looks like, but I think it might make it a bit more readable if you used the change_to_susceptible instead of ~sick and change_to_diseased instead of sick to update the self.cause.name columns as well. as it is, we update the disease column every time step but event time columns only if the state changes. It seems better to just do both only if the state changes.

PR #149, opened 2019-05-17
metrics names

halfhorst:
Names in the metrics subfolder


src/vivarium_public_health/metrics/risk.py
```diff
@@ -45,6 +45,10 @@ def __init__(self, risk: str):
             f'{self.risk.name}_observer': CategoricalRiskObserver.configuration_defaults['metrics']['risk_observer']
         }}
 
+    @property
```
kfwilson: do we want a repr on this class like the other observers?

src/vivarium_public_health/metrics/treatment.py
```diff
@@ -42,11 +42,14 @@ class TreatmentObserver:
 
     def __init__(self, treatment: str):
         self.treatment = treatment
-        self.name = f'{self.treatment}_observer'
         self.configuration_defaults = {
             'metrics': {self.name: TreatmentObserver.configuration_defaults['metrics']['treatment_observer']}
         }
 
+    @property
```
kfwilson: do we want a repr for this class too?

src/vivarium_public_health/metrics/inspector.py
```diff
@@ -2,6 +2,11 @@
 
 
 class Inspector:
+
```
collijk: Delete this component

PR #150, opened 2019-05-17
artifact manager names

halfhorst:


halfhorst:
> we put reprs on the managers in vivarium - do we want to do that here too?

may as well, while we're here

PR #152, opened 2019-05-17
disease names

halfhorst:
Add names behind props, add some reprs where missing so printing works ok

### DIFF
--- a/src/vivarium_public_health/disease/model.py
+++ b/src/vivarium_public_health/disease/model.py
@@ -16,9 +16,9 @@ class DiseaseModelError(VivariumError):
 
 class DiseaseModel(Machine):
     def __init__(self, cause, initial_state=None, get_data_functions=None, cause_type="cause", **kwargs):
+        super().__init__(cause, **kwargs)
         self.cause = cause
         self.cause_type = cause_type
-        super().__init__(cause, **kwargs)
 
         if initial_state is not None:
             self.initial_state = initial_state.state_id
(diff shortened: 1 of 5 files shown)


src/vivarium_public_health/disease/model.py
```diff
@@ -31,9 +31,13 @@ def __init__(self, cause, initial_state=None, get_data_functions=None, cause_typ
...
                 f"{self.cause_type}.{cause}.cause_specific_mortality")
 
+    @property
+    def name(self):
+        return f"disease_model.{self.cause}"
+
     @property
     def condition(self):
-        return self.state_column
+        return self.cause
```
halfhorst: these are synonyms but I felt it was confusing to rely on the attribute of super, this was a little more clear to me

kfwilson: hmm I agree it was confusing. The only thing I'm a little worried about is technically self.cause and self.state_column are synonyms only because we both assign them to equal cause in init. If we ever change that, it could be problematic, because condition is used below as the creates_columns arg to initializes_simulants (line 56). But it does seem unnecessary to have state_column, cause, and condition all now be the same thing.

halfhorst: yeah I'd like to close this loop. It seems like condition should just mean cause b/c I don't know the intent behind making this property in the first place throughout this whole PR i've wanted to try and limit reaching into the super class where not totally necessary to make the inheritance less gnarly

PR #153, opened 2019-05-17
risk names

halfhorst:


src/vivarium_public_health/risks/distributions.py
```diff
@@ -84,12 +105,22 @@ def ppf(self, x):
         category_index = (exposure_sum.T < x).T.sum('columns')
         return pd.Series(np.array(self.categories)[category_index], name=self.risk + '_exposure', index=x.index)
 
+    def __str__(self):
+        return f"PolytomousDistribution(risk= {self.risk}, categories= {self.categories}"
+
+    def __repr__(self):
```
kfwilson: These reprs for the distributions seem a little inconsistent - sometimes they include e.g. exposure_data, means, sd, etc. just as strings, sometimes they include the data, sometimes they aren't included at all.

src/vivarium_public_health/risks/distributions.py
```diff
@@ -63,6 +77,9 @@ def ppf(self, q):
             x = pd.Series([])
         return x
 
+    def __repr__(self):
+        return f"SimulationDistribution(risk={self.risk}, mean={self.mean}, sd={self.sd})"
```
kfwilson: I don't think there is a self.mean or self.sd for this class?

PR #154, opened 2019-05-17
population names

halfhorst:


src/vivarium_public_health/population/add_new_birth_cohorts.py
```diff
@@ -48,6 +52,12 @@ def on_time_step(self, event):
                                       'sim_state': 'time_step',
                                   })
 
+    def __str__(self):
```
kfwilson: why do we have a str and a repr for this fertility component but only a repr for the others?

halfhorst: i've been staring at too many things :eyes:

PR #160, opened 2019-07-12
Feature/age filter metrics

kfwilson:
Previously, observers if stratified by age only reported age groups up to exit_age. This makes it so that the age groups are also bounded on the lower edge - age_groups below the specified age_start will no longer be included. Also added a test to show that both the age_start and exit_age restrictions work and that the age_group_end of the oldest age group is capped to exit_age if it straddles the exit_age

### DIFF
--- a/src/vivarium_public_health/metrics/utilities.py
+++ b/src/vivarium_public_health/metrics/utilities.py
@@ -142,6 +142,7 @@ def get_age_bins(builder) -> pd.DataFrame:
     age_start = builder.configuration.population.age_start
     min_bin_start = age_bins.age_group_start[np.asscalar(np.digitize(age_start, age_bins.age_group_end))]
     age_bins = age_bins[age_bins.age_group_start >= min_bin_start]
+    age_bins.loc[age_bins.age_group_start < age_start, 'age_group_start'] = age_start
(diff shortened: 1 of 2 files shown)


src/vivarium_public_health/metrics/utilities.py
```diff
@@ -137,6 +137,12 @@ def get_age_bins(builder) -> pd.DataFrame:
 
     """
     age_bins = builder.data.load('population.age_bins')
+
+    # Works based on the fact that currently only models with age_start = 0 can include fertility
+    age_start = builder.configuration.population.age_start
+    min_bin_start = age_bins.age_group_start[np.asscalar(np.digitize(age_start, age_bins.age_group_end))]
+    age_bins = age_bins[age_bins.age_group_start >= min_bin_start]
+
```
collijk: You also want to set the 'age_group_start' on the lower bin if necessary.  E.g. if we're using 5 year age bins in the data and the user specifies an age start of 32, the bin corresponding to the lowest age should be 32-35.

kfwilson: Good point. Just added that + updated test to verify

PR #161, opened 2019-07-16
Clip tails of simulation distribution to match behavior of ensemble d…

collijk:
…istribution

### DIFF
--- a/src/vivarium_public_health/risks/distributions.py
+++ b/src/vivarium_public_health/risks/distributions.py
@@ -43,10 +43,7 @@ class EnsembleSimulation:
 
     def ppf(self, q):
         if not q.empty:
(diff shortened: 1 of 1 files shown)


src/vivarium_public_health/risks/distributions.py
```diff
@@ -80,6 +80,8 @@ def _get_parameters(self, mean, sd):
 
     def ppf(self, q):
         if not q.empty:
+            q[q >= 0.999] = 0.998
```
aflaxman: 1) Got a reference for where these magic numbers are found in GBD?
2) Are you sure you want to have a non-monotonic mapping where, for example, although 0.9989 < 0.9990 we have f(0.9989) > f(0.9990) ?

PR #162, opened 2019-07-24
Mic966 intervention start

KjellSwedin:
Add "intervention_start_date" to vph v0.8.

Is there a sensible default for mass_treatment_campaign.py configuration_defaults? Currently set to 1/1/2030

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ if __name__ == "__main__":
         long_description = f.read()
 
     install_requirements = [
-        'vivarium<0.8.20',
+        'vivarium==0.8.20',
         # FIXME: Newer versions of numpy have conflicting dependencies
         # with pytables.
         'numpy<=1.15.4',
--- a/setup.py
+++ b/setup.py
@@ -17,11 +17,11 @@ if __name__ == "__main__":
         long_description = f.read()
 
     install_requirements = [
-        'vivarium>=0.8.19',
+        'vivarium<0.8.20',
         # FIXME: Newer versions of numpy have conflicting dependencies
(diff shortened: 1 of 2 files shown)


src/vivarium_public_health/treatment/mass_treatment_campaign.py
```diff
@@ -63,6 +67,10 @@ def setup(self, builder):
         self.config = builder.configuration[self.treatment_name]
         self.clock = builder.time.clock()
 
+        self.start_date = pd.Timestamp(**builder.configuration.time['start'].to_dict())
```
collijk: I don't think you need to hang on to the start and end dates.

setup.py
```diff
@@ -17,11 +17,11 @@
         long_description = f.read()
 
     install_requirements = [
-        'vivarium>=0.8.19',
+        'vivarium<0.8.20',
```
collijk: Probably need a hard pin here.  Don't think this will work with 0.8.18

setup.py
```diff
@@ -17,11 +17,11 @@
         long_description = f.read()
 
     install_requirements = [
-        'vivarium>=0.8.19',
+        'vivarium==0.8.20',
         # FIXME: Newer versions of numpy have conflicting dependencies
         # with pytables.
         'numpy<=1.15.4',
-        'pandas',
+        'pandas<0.25',
```
kfwilson: I'm not sure we want to pin pandas here actually. The pin is because of a central comp dependency that only affects data so it doesn't seem like it needs to be pinned here

PR #166, opened 2019-08-21
Update event emitter args

collijk:
Bring Kjell's upstream changes into public health


collijk:
Checks are failing because it is using develop instead of the latest release branch

PR #177, opened 2019-09-16
Feature/dependency manager

collijk:
Since I had to go module by module, I also made some ordering/naming changes in the components.  In particular, population initializers are now always methods named `on_initialize_simulants`, event handlers are now always named `on_{event}`, and within a component, method ordering looks like:
```
__init__
name
sub_components
{other properties}
setup
on_initialize_simulants
{event handlers in order of execution}
pipeline sources
pipeline modifiers
helper methods
metrics
__repr__/__str__
```

### DIFF
--- a/src/vivarium_public_health/disease/model.py
+++ b/src/vivarium_public_health/disease/model.py
@@ -50,82 +50,23 @@ class DiseaseModel(Machine):
         cause_specific_mortality_rate = self.load_cause_specific_mortality_rate_data(builder)
         self.cause_specific_mortality_rate = builder.lookup.build_table(cause_specific_mortality_rate)
         builder.value.register_value_modifier('cause_specific_mortality_rate',
-                                              self.adjust_cause_specific_mortality_rate)
+                                              self.adjust_cause_specific_mortality_rate,
+                                              requires_columns=['age', 'sex'])
 
         builder.value.register_value_modifier('metrics', modifier=self.metrics)
 
         self.population_view = builder.population.get_view(['age', 'sex', self.state_column])
-        builder.population.initializes_simulants(self.load_population_columns,
+        builder.population.initializes_simulants(self.on_initialize_simulants,
                                                  creates_columns=[self.state_column],
-                                                 requires_columns=['age', 'sex'])
-        self.randomness = builder.randomness.get_stream('{}_initial_states'.format(self.state_column))
+                                                 requires_columns=['age', 'sex'],
+                                                 requires_streams=[f'{self.state_column}_initial_states'])
+        self.randomness = builder.randomness.get_stream(f'{self.state_column}_initial_states')
 
-        self._propensity = pd.Series()
-        self.propensity = builder.value.register_value_producer(f'{self.cause}.prevalence_propensity',
-                                                                source=lambda index: self._propensity[index])
+        builder.event.register_listener('time_step', self.on_time_step)
+        builder.event.register_listener('time_step__cleanup', self.on_time_step_cleanup)
 
-        builder.event.register_listener('time_step', self.time_step_handler)
-        builder.event.register_listener('time_step__cleanup', self.time_step__cleanup_handler)
-
-    def load_cause_specific_mortality_rate_data(self, builder):
-        if 'cause_specific_mortality_rate' not in self._get_data_functions:
-            only_morbid = builder.data.load(f'cause.{self.cause}.restrictions')['yld_only']
-            if only_morbid:
-                csmr_data = 0
-            else:
-                csmr_data = builder.data.load(f"{self.cause_type}.{self.cause}.cause_specific_mortality_rate")
-        else:
-            csmr_data = self._get_data_functions['cause_specific_mortality_rate'](self.cause, builder)
-        return csmr_data
-
-    def adjust_cause_specific_mortality_rate(self, index, rate):
-        return rate + self.cause_specific_mortality_rate(index)
-
-    def _get_default_initial_state(self):
-        susceptible_states = [s for s in self.states if isinstance(s, SusceptibleState)]
-        if len(susceptible_states) != 1:
-            raise DiseaseModelError("Disease model must have exactly one SusceptibleState.")
-        return susceptible_states[0].state_id
-
-    def time_step_handler(self, event):
-        self.transition(event.index, event.time)
-
-    def time_step__cleanup_handler(self, event):
-        self.cleanup(event.index, event.time)
-
-    def get_state_weights(self, pop_index, prevalence_type):
-        states = [s for s in self.states
-                  if hasattr(s, f'{prevalence_type}_data') and getattr(s, f'{prevalence_type}_data') is not None]
-
-        if not states:
-            return states, None
-
-        weights = [getattr(s, f'{prevalence_type}_data')(pop_index) for s in states]
-        for w in weights:
-            w.reset_index(inplace=True, drop=True)
-        weights += ((1 - np.sum(weights, axis=0)), )
-
-        weights = np.array(weights).T
-        weights_bins = np.cumsum(weights, axis=1)
-
-        state_names = [s.state_id for s in states] + [self.initial_state]
-
-        return state_names, weights_bins
-
-    @staticmethod
-    def assign_initial_status_to_simulants(simulants_df, state_names, weights_bins, propensities):
-        simulants = simulants_df[['age', 'sex']].copy()
-
-        choice_index = (propensities.values[np.newaxis].T > weights_bins).sum(axis=1)
-        initial_states = pd.Series(np.array(state_names)[choice_index], index=simulants.index)
-
-        simulants.loc[:, 'condition_state'] = initial_states
-        return simulants
-
-    def load_population_columns(self, pop_data):
+    def on_initialize_simulants(self, pop_data):
         population = self.population_view.subview(['age', 'sex']).get(pop_data.index)
-
-        self._propensity = self._propensity.append(self.randomness.get_draw(pop_data.index))
 
         assert self.initial_state in {s.state_id for s in self.states}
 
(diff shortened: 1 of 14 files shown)


src/vivarium_public_health/disease/model.py
```diff
@@ -50,22 +50,59 @@ def setup(self, builder):
...
 
-        self._propensity = pd.Series()
-        self.propensity = builder.value.register_value_producer(f'{self.cause}.prevalence_propensity',
-                                                                source=lambda index: self._propensity[index])
+        builder.event.register_listener('time_step', self.on_time_step)
+        builder.event.register_listener('time_step__cleanup', self.on_time_step_cleanup)
 
-        builder.event.register_listener('time_step', self.time_step_handler)
-        builder.event.register_listener('time_step__cleanup', self.time_step__cleanup_handler)
+    def on_initialize_simulants(self, pop_data):
```
collijk: This whole block is a move and rename of methods

src/vivarium_public_health/disease/model.py
```diff
@@ -87,20 +124,14 @@ def _get_default_initial_state(self):
...
-    def time_step_handler(self, event):
-        self.transition(event.index, event.time)
-
-    def time_step__cleanup_handler(self, event):
-        self.cleanup(event.index, event.time)
-
     def get_state_weights(self, pop_index, prevalence_type):
         states = [s for s in self.states
-                  if hasattr(s, f'{prevalence_type}_data') and getattr(s, f'{prevalence_type}_data') is not None]
+                  if hasattr(s, f'{prevalence_type}') and getattr(s, f'{prevalence_type}') is not None]
```
collijk: Actually important.  I removed the 'data' suffix from the prevalence look up tables since it was not totally true.

src/vivarium_public_health/disease/special_disease.py
```diff
@@ -11,6 +11,7 @@
 import re
 
 import pandas as pd
+from vivarium.framework.values import list_combiner, joint_value_post_processor
```
collijk: In this file, and in state.py, I was having a hard time parsing through the data/dependency structures.  They are now reorganized and essentially the same blocks of code (indicating our abstractions are not so great; no surprise)

collijk: This is all around loading data to set up look up tables and pipelines.

Note, none of the test code needed to change in order to make the test pass.

collijk: Generically, the pattern is
```
load data (checking conditions as necessary, shove this off to a method that takes the builder as an arg when messy)
turn data into lookup table
use lookup table as source/modifier for appropriate pipelines
``

Comment (collijk) [reply src/vivarium_public_health/disease/special_disease.py]: Structurally, riskattributable diseases will now always have mortality data/pipelines, they may just be sourced with all-zero data if appropriate.

File: src/vivarium_public_health/disease/special_disease.py [hunk 4]
```diff
@@ -155,6 +177,42 @@ def on_time_step(self, event):
 
         self.population_view.update(pop)
 
+    def compute_disability_weight(self, index):
```
collijk: These are moved and renamed.  Behavior has changed as well.  Now only living people will have non-zero disability/excess mortality rates.  This should prevent broad categories of errors we see periodically.

src/vivarium_public_health/disease/state.py
```diff
@@ -39,16 +39,26 @@ def setup(self, builder):
 
         self.clock = builder.time.clock()
 
-        columns = [self._model, 'alive']
-        columns += [self.event_time_column, self.event_count_column]
-
-        self.population_view = builder.population.get_view(columns)
-        builder.population.initializes_simulants(self.load_population_columns,
-                                                 creates_columns=[self.event_time_column, self.event_count_column],
+        columns_created = [self.event_time_column, self.event_count_column]
```
collijk: notes from special disease apply here as well.  You'll find except for a few names, most of the code is identical.

src/vivarium_public_health/disease/state.py
```diff
@@ -302,81 +346,56 @@ def _cleanup_effect(self, index, event_time):
...
-    def __repr__(self):
-        return 'TransientDiseaseState(name={})'.format(self.state_id)
+        if isinstance(dwell_time, pd.Timedelta):
+            dwell_time = dwell_time.total_seconds() / (60 * 60 * 24)
+        if (isinstance(dwell_time, pd.DataFrame) and np.any(dwell_time.value != 0)) or dwell_time > 0:
+            self.transition_set.allow_null_transition = True
 
+        return dwell_time
 
-class MortalityEffect:
```
collijk: Like the special disease, all disease states now have csmr/excess mortality.  It is appropriately set to zero when necessary.

src/vivarium_public_health/disease/transition.py
```diff
@@ -19,16 +19,28 @@ def __init__(self, input_state, output_state, get_data_functions=None, **kwargs)
...
         self.joint_paf = builder.value.register_value_producer(f'{pipeline_name}.paf',
                                                                source=lambda index: [paf(index)],
                                                                preferred_combiner=list_combiner,
                                                                preferred_post_processor=joint_value_post_processor)
 
-    def _get_rate_data(self, builder):
+        self.population_view = builder.population.get_view(['alive'])
+
+    def compute_transition_rate(self, index):
+        transition_rate = pd.Series(0, index=index)
```
collijk: Dead people no longer have transition rates.

src/vivarium_public_health/population/add_new_birth_cohorts.py
```diff
@@ -158,36 +156,38 @@ def setup(self, builder):
         builder : vivarium.engine.Builder
             Framework coordination object.
         """
+        age_specific_fertility_rate = self.load_age_specific_fertility_rate_data(builder)
```
collijk: This whole component needs serious attention before we put it in a production model.  I wrote it when I didn't understand demography/our sim conventions super well and it will be mildly incorrect and confusing to use.

src/vivarium_public_health/population/mortality.py
```diff
@@ -63,5 +65,11 @@ def mortality_handler(self, event):
             dead_pop['years_of_life_lost'] = self.life_expectancy(dead_pop.index)
             self.population_view.update(dead_pop[['alive', 'exit_time', 'cause_of_death', 'years_of_life_lost']])
 
+    def calculate_mortality_rate(self, index):
+        acmr = self.all_cause_mortality_rate(index)
+        csmr = self.cause_specific_mortality_rate(index, skip_post_processor=True)
+        cause_deleted_mortality_rate = acmr - csmr
+        return pd.DataFrame({'other_causes': cause_deleted_mortality_rate})
```
collijk: Note, this is *always* a dataframe now, which let me clean up some lines of code elsewhere in the disease stuff that had to handle type conversions.

collijk: Also, we no longer have a mix of ``death_due_to_other_causes`` and cause names in the ``cause_of_death`` column.

tests/disease/test_special_disease.py
```diff
@@ -70,25 +70,6 @@ def test_filter_by_exposure_continuous(disease_mock, distribution, threshold):
     assert np.all(expected(test_index) == filter_func(test_index))
 
 
-def test_mortality_rate_pandas_series(disease_mock):
```
collijk: This test no longer applies.

collijk:
I mis-named the branches which is causing CI to fail, but test do pass against the `feature/dependency-interface` branch of vivarium.

src/vivarium_public_health/disease/transition.py
```diff
@@ -19,16 +19,28 @@ def __init__(self, input_state, output_state, get_data_functions=None, **kwargs)
         self._get_data_functions = get_data_functions if get_data_functions is not None else {}
 
     def setup(self, builder):
-        rate_data, pipeline_name = self._get_rate_data(builder)
+        rate_data, pipeline_name = self.load_transition_rate_data(builder)
         self.base_rate = builder.lookup.build_table(rate_data)
-        self.effective_rate = builder.value.register_rate_producer(pipeline_name, source=self.rates)
+        self.transition_rate = builder.value.register_rate_producer(pipeline_name,
+                                                                    source=self.compute_transition_rate,
+                                                                    requires_columns=['age', 'sex', 'alive'])
```
kfwilson: this also needs `requires_values=['f{pipeline_name}.paf']`

src/vivarium_public_health/risks/effect.py
```diff
@@ -77,9 +78,12 @@ def setup(self, builder):
 
         self.exposure_effect = get_exposure_effect(builder, self.risk)
 
-        builder.value.register_value_modifier(f'{self.target.name}.{self.target.measure}', modifier=self.adjust_target)
+        builder.value.register_value_modifier(f'{self.target.name}.{self.target.measure}',
+                                              modifier=self.adjust_target,
+                                              requires_values=[f'{self.risk.name}.exposure'])
```
kfwilson: I think this also needs `requires_columns=['age', 'sex']` because it uses the rr lookup table

collijk: Yeah, I now don't like how hidden all that is.

src/vivarium_public_health/disease/special_disease.py
```diff
@@ -190,19 +248,17 @@ def continuous_filter(index):
...
+        if builder.configuration[self.cause.name].mortality:
+            emr_data = builder.data.load(f'cause.{self.cause.name}.excess_mortality_rate')
+        else:
+            emr_data = 0
+        return emr_data
 
-    def compute_disability_weight(self, index):
-        disability_weight = pd.Series(0, index=index)
-        with_condition = self.population_view.get(index,
-                                                  query=f'alive=="alive" and {self.cause.name}=="{self.cause.name}"')
```
halfhorst: didn't we limit dw to only the living before, too?

PR #180, opened 2019-09-24
Feature/update lookup cols

kfwilson:
update lookup table parameter columns to match vivarium updates

### DIFF
--- a/src/vivarium_public_health/population/add_new_birth_cohorts.py
+++ b/src/vivarium_public_health/population/add_new_birth_cohorts.py
@@ -157,8 +157,7 @@ class FertilityAgeSpecificRates:
             Framework coordination object.
         """
         age_specific_fertility_rate = self.load_age_specific_fertility_rate_data(builder)
-        fertility_rate = builder.lookup.build_table(age_specific_fertility_rate, key_columns=(),
-                                                    parameter_columns=['age', 'year'])
+        fertility_rate = builder.lookup.build_table(age_specific_fertility_rate, parameter_columns=['age', 'year'])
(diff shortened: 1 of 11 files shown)


src/vivarium_public_health/population/mortality.py
```diff
@@ -31,9 +32,8 @@ def setup(self, builder):
                                                                    requires_columns=['age', 'sex'])
 
         life_expectancy_data = builder.data.load("population.theoretical_minimum_risk_life_expectancy")
-        self.life_expectancy = builder.lookup.build_table(life_expectancy_data,
-                                                          key_columns=[],
-                                                          parameter_columns=[('age', 'age_start', 'age_end')])
+        self.life_expectancy = builder.lookup.build_table(life_expectancy_data, key_columns=[],
```
collijk: Can you drop the key cols here?

src/vivarium_public_health/population/add_new_birth_cohorts.py
```diff
@@ -157,7 +157,8 @@ def setup(self, builder):
             Framework coordination object.
         """
         age_specific_fertility_rate = self.load_age_specific_fertility_rate_data(builder)
-        fertility_rate = builder.lookup.build_table(age_specific_fertility_rate, key_columns=())
+        fertility_rate = builder.lookup.build_table(age_specific_fertility_rate, key_columns=(),
```
collijk: Do you need key cols, if so, provide a list instead of a tuple for consistency.

kfwilson: made a small change in vivarium so that key cols can be None (the default if not provided) so I dropped these

PR #182, opened 2019-09-25
Bugfix/mslt tables

pandersen13:
These are the changes required for MSLT simulations to run.

### DIFF
--- a/src/vivarium_public_health/mslt/delay.py
+++ b/src/vivarium_public_health/mslt/delay.py
@@ -399,8 +399,8 @@ class DelayedRisk:
             # causes additional remission.
             tax_inc = self.tax_effect_inc(idx)
             tax_rem = self.tax_effect_rem(idx)
-            int_inc *= tax_inc
-            int_rem += (1 - tax_rem) * pop[col_int_yes]
+            int_inc = int_inc * tax_inc
+            int_rem = int_rem + (1 - tax_rem) * pop[col_int_yes]
 
         # Apply the incidence rate to the never-exposed population.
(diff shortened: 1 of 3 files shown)


src/vivarium_public_health/mslt/delay.py
```diff
@@ -477,3 +477,18 @@ def incidence_adjustment(self, disease, index, incidence_rate):
...
+    """Helper method for loading dataframe from artifact.
+
+    Performs a long to wide conversion if dataframe has an index column
+    named 'measure'.
+
+    """
+    data = builder.data.load(entity_key)
+
+    if 'measure' in data.columns :
+        data  = data.pivot_table(index = [i for i in data.columns if i not in ['measure','value']], columns = 'measure', \
```
collijk: Pivot table is actually quite slow, if I recall.  You might get better performance using dataframe.unstack() and then fixing up column/index names.  

Style nitpicks: when you're giving keyword args, no spaces around the '=' operator.  You also should stay away from explicit line continuation as recommended by pep8 (https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

When doing method chaining in pandas, I find the following format easiest to read:
```
index = [i for i in data.columns if i not in ['measure','value']]
data = (
    data.pivot_table(index=index, columns='measure', values='value)
    .rename_axis(None, axis=1)
    .reset_index()
)
```
This is especially helpful if your method chains get quite long.  The most important code optimization is for readability.

collijk:
I'll probably merge this in later today and make a release you can pin to in the tobacco repository.

PR #186, opened 2019-11-27
fix disease observer configuration key

halfhorst:
the disease observer was using its name as its configuration key, which doesn't work


collijk:
PR into master, make a minor release bump, and merge master into develop

PR #188, opened 2019-12-13
This looks like a regression issue -- this worked in the conic_sam pr…

KjellSwedin:
…oject but failed when I tried it in the conic_lsff project


collijk:
Definitely, I went through and dramatically reduced the boilerplate in that module during the last big release.  Rebase this against master and make a patch release.

collijk:
Then merge into develop.

PR #194, opened 2020-01-29
Change 'remission' to 'value'

pandersen13:
Extra 'remission' column was being created instead of 'value' column being overwritten. This resulted in a table of NaNs when multiplied by a pop column.


robmoss:
Good work Patrick.

To avoid finding or making these kinds of errors in the future, it's probably worth checking that data frames don't have any NA values (a) when they're first loaded; and (b) whenever we manipulate them. Something like:

```
if dataframe.isna().values.any():
    raise ValueError("useful message goes here")
```

What do you think?

PR #197, opened 2020-02-12
call get_seed from randomness interface

rmudambi:
Tests will fail until the associated PR for `vivarium` (https://github.com/ihmeuw/vivarium/pull/150) is merged.


src/vivarium_public_health/population/add_new_birth_cohorts.py
```diff
@@ -106,7 +106,7 @@ def setup(self, builder):
         self.birth_rate = get_live_births_per_year(builder)
 
         self.clock = builder.time.clock()
-        self.randomness = builder.randomness.get_stream('crude_birth_rate')
```
collijk: I'm not a giant fan of hanging on to the builder interface rather than requesting the required resource.  Maybe we do a thing like we do with the simulation clock:  have `get_seed` return a callable that produces a unique seed each time step.  

A pattern that is almost true:
Every call to the builder looks like `builder.SERVICE.register_X`, takes a callable and other info, and returns `None` or looks like `builder.SERVICE.get_X` and returns a callable.  

I think let's do what you have for now.  I've definitely broken that pattern in newer managers.  Let's talk together about whether that's a pattern worth enforcing, and then we'll make some tickets to clean up places where it's not consistent if we decide to.

collijk:
Merge the upstream PR, then retrigger the CI so we can verify the build passes.  If you duplicate your branch names across the upstream dependencies, travis will find them and run the test suite against them.

src/vivarium_public_health/population/add_new_birth_cohorts.py
collijk: Ticket: https://jira.ihme.washington.edu/browse/MIC-1613

PR #199, opened 2020-02-27
update vivarium pin to latest, unpin pandas and numpy

halfhorst:
Two tests fail that use randomness's updated interface with get_seed, which is still on develop. This release must be coordinated with that release


halfhorst:
on hold while I investigate tables 3.4.4

collijk:
@halfhorst Should we close this?

PR #201, opened 2020-02-28
refactor final age bin rescale to simplify and remove warning

halfhorst:
This function is responsible for excluding simulants out of scope due to exit_age and scaling the value of simulants who cross the exit_age threshold. I refactored it just a bit to remove SettingWithCopyWarnings


src/vivarium_public_health/population/data_transformations.py
```diff
@@ -417,12 +417,12 @@ def get_live_births_per_year(builder):
 def rescale_final_age_bin(builder, population_data):
```
collijk: Make math clearer with local variables.

PR #202, opened 2020-02-29
remove clean_cause_of_death

halfhorst:
This is now performed by get_deaths


halfhorst:
upon further examination, this is not taken care of

PR #205, opened 2020-03-11
Feature/results manager

collijk:
See https://github.com/ihmeuw/vivarium/pull/154 first for context.

### DIFF
--- a/src/vivarium_public_health/metrics/metrics.py
+++ b/src/vivarium_public_health/metrics/metrics.py
@@ -1,0 +1,53 @@
+import typing
+
+import numpy as np
+
+if typing.TYPE_CHECKING:
+    from vivarium.framework.engine import Builder
+
+
+class Metrics:
+
+    configuration_defaults = {
+        'metrics': {
+            'stratification_params': []
+        }
+    }
+
+    @property
+    def name(self):
+        # Fixme: Avoiding name collision for now.
+        return 'metrics2'
+
+    def setup(self, builder: 'Builder'):
+        self.config = builder.configuration.metrics
+        if 'age_group' in self.config.stratification_params:
+            age_bins = self.get_age_bins(builder)
+            target_col = 'age'
+            result_col = 'age_group'
+            bins = age_bins.age_start.to_list() + [age_bins.age_end.iloc[-1]]
+            labels = age_bins.age_group_name.str.replace(' ', '_').str.lower().to_list()
+            builder.results.add_binner(target_col, result_col, bins, labels, include_lowest=True)
+        if 'year' in self.config.stratification_params:
+            for time_type in ['current', 'event', 'entrance', 'exit']:
+                builder.results.add_mapper(f'{time_type}_time', f'{time_type}_year',
+                                           lambda x: x.dt.year, is_vectorized=True)
+
+    def get_age_bins(self, builder: 'Builder'):
+        age_bins = builder.data.load('population.age_bins')
+
+        # Works based on the fact that currently only models with age_start = 0 can include fertility
+        age_start = builder.configuration.population.age_start
+        min_bin_start = age_bins.age_start[np.asscalar(np.digitize(age_start, age_bins.age_end))]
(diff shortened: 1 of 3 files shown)


src/vivarium_public_health/metrics/metrics.py
```diff
@@ -0,0 +1,57 @@
...
+            labels = age_bins.age_group_name.str.replace(' ', '_').str.lower().to_list()
+            builder.results.add_binner(target_col, result_col, bins, labels, include_lowest=True)
+        # Map the datetime columns into int columns for year stratification.
+        if 'year' in self.config.stratification_params:
+            for time_type in ['current', 'event', 'entrance', 'exit']:
+                builder.results.add_mapper(f'{time_type}_time', f'{time_type}_year',
+                                           lambda x: x.dt.year, is_vectorized=True)
+
+    @staticmethod
+    def get_age_bins(builder: 'Builder'):
```
collijk: Pulled directly from the normal metrics utilities.  should only be needed here in the future.

src/vivarium_public_health/metrics/new_disability.py
```diff
@@ -0,0 +1,102 @@
...
+
+        cause_states = builder.components.get_components_by_type(tuple(self.disease_classes))
+        cause_names = [c.state_id for c in cause_states]
+        self.disability_weight_pipelines = {cause: builder.value.get_value(f'{cause}.disability_weight')
+                                            for cause in cause_names}
+
+        params = builder.configuration.metrics.stratification_params
+        groupers = [] + self.additional_groupers
+        columns = ['tracked', 'alive'] + self.additional_columns
+        if 'age_group' in params:
```
collijk: Still thinking about how to get rid of this block; it'll show up in every observer for now.

collijk: The issue is mainly around year, which requires a different grouping for different observers (e.g. exit year for deaths)

collijk: Actually, strike that.  we'll probably just collect deaths each time step instead.

src/vivarium_public_health/metrics/new_disability.py
```diff
@@ -0,0 +1,102 @@
...
+            columns.append('sex')
+        if 'year' in params:
+            groupers.append('current_year')
+
+        for cause in cause_states:
+            name = cause.state_id
+            pipeline = builder.value.get_value(f'{name}.disability_weight')
+            builder.results.register_results_producer(
+                measure=f'ylds_due_to_{name}',
+                pop_filter='alive == "alive"',
```
collijk: tracked == True

src/vivarium_public_health/metrics/new_disability.py
```diff
@@ -0,0 +1,102 @@
...
+        if 'age_group' in params:
+            groupers.append('age_group')
+            columns.append('age')
+        if 'sex' in params:
+            groupers.append('sex')
+            columns.append('sex')
+        if 'year' in params:
+            groupers.append('current_year')
+
+        for cause in cause_states:
```
collijk: for cause_state in cause_states

src/vivarium_public_health/metrics/new_disability.py
```diff
@@ -0,0 +1,102 @@
...
+                pop_filter='alive == "alive"',
+                groupers=groupers,
+                aggregator=_dw_aggregator(pipeline),
+                requires_columns=columns,
+                when='time_step__prepare',
+            )
+        builder.results.register_results_producer(
+            measure='ylds_due_to_all_causes',
+            pop_filter='alive == "alive"',
+            groupers=groupers,
```
collijk: groupby_cols

src/vivarium_public_health/metrics/new_disability.py
```diff
@@ -0,0 +1,102 @@
...
+            aggregator=_dw_aggregator(self.disability_weight),
```
collijk: Consider using the name to refference groupby(cols).apply(aggregator)

PR #212, opened 2020-12-28
Feature/GitHub actions ci

KjellSwedin:


.github/workflows/build.yml
```diff
@@ -0,0 +1,63 @@
+# -----------------------------------------------------------------------------
```
rmudambi, line 1: don't we need a deploy.yml file?

mattkappel (approved):

rmudambi (approved):

.github/workflows/build.yml
KjellSwedin: I will do it as a follow-on PR

PR #217, opened 2021-01-26
speed up ppf calculation in polytomous risks

rmudambi:
verified that the new code generates the same output as the old.

### DIFF
--- a/src/vivarium_public_health/risks/distributions.py
+++ b/src/vivarium_public_health/risks/distributions.py
@@ -137,7 +137,7 @@ class PolytomousDistribution:
         if not np.allclose(1, np.sum(sorted_exposures, axis=1)):
             raise MissingDataError('All exposure data returned as 0.')
         exposure_sum = sorted_exposures.cumsum(axis='columns')
-        category_index = (exposure_sum.T < x).T.sum('columns')
+        category_index = pd.concat([exposure_sum[c] < x for c in exposure_sum.columns], axis=1).sum(axis=1)
         return pd.Series(np.array(self.categories)[category_index], name=self.risk + '_exposure', index=x.index)
 
     def __repr__(self):


src/vivarium_public_health/risks/distributions.py
```diff
@@ -137,7 +137,7 @@ def ppf(self, x):
         if not np.allclose(1, np.sum(sorted_exposures, axis=1)):
             raise MissingDataError('All exposure data returned as 0.')
         exposure_sum = sorted_exposures.cumsum(axis='columns')
-        category_index = (exposure_sum.T < x).T.sum('columns')
+        category_index = pd.concat([exposure_sum[c] < x for c in exposure_sum.columns], axis=1).sum(axis=1)
```
collijk, line 140: Cool.  You could also use np.sum([...], axis=1) and dodge the concat overhead.  Though that's pretty trivial here compared to what we had before.

collijk (approved):

src/vivarium_public_health/risks/distributions.py
rmudambi: Just tested it, and it's actually `np.sum([...], axis=0)`, but it also seems to take about 2x as much time as the `pd.concat` apporach.

src/vivarium_public_health/risks/distributions.py
```diff
@@ -137,7 +137,7 @@ def ppf(self, x):
         if not np.allclose(1, np.sum(sorted_exposures, axis=1)):
             raise MissingDataError('All exposure data returned as 0.')
         exposure_sum = sorted_exposures.cumsum(axis='columns')
```
aflaxman, line 139: Would it perhaps be even faster if you calculate `exposure_sum` once in `setup` instead of recomputing it in every call to `ppf`? IIUC, it does not change after setup.

PR #221, opened 2021-02-22
Feature/state transition strings

KjellSwedin:


mattkappel (approved):

src/vivarium_public_health/disease/special_disease.py
```diff
@@ -94,13 +98,25 @@ def __init__(self, cause, risk):
         self.configuration_defaults = {
             self.cause.name: RiskAttributableDisease.configuration_defaults['risk_attributable_disease']
         }
+        self._state_names = [f'{self.cause.name}', f'susceptible_to_{self.cause.name}']
+        self._transition_names = [f'susceptible_to_{self.cause.name}_TO_{self.cause.name}',
+                                    f'{self.cause.name}_TO_susceptible_to_{self.cause.name}']
```
collijk, line 103: The second transition here should only exist if the disease is recoverable, which you can find in the configuration

src/vivarium_public_health/disease/special_disease.py
```diff
@@ -253,6 +269,11 @@ def continuous_filter(index):
 
         return filter_function
 
+    def adjust_state_and_transitions(self):
```
collijk, line 272: Ah, see you're doing a thing here.  If the disease is recoverable, you have an SIS model, not an SIR model.  If the disease is not recoverable you have an SI model.

collijk (changes requested):
Address the comments about SIS vs SIR and then everything else looks good.

src/vivarium_public_health/disease/special_disease.py
```diff
@@ -94,13 +98,25 @@ def __init__(self, cause, risk):
...
+                                    f'{self.cause.name}_TO_susceptible_to_{self.cause.name}']
 
     @property
     def name(self):
-        return f'risk_attributable_disease.{self.cause}.{self.risk}'
+        return f'disease_model.{self.cause.name}'
+
+    @property
+    def state_names(self):
+        return self._state_names
```
collijk, line 111: Optional: you could return the list literals here as well (e.g. generate them when they're called since they're super cheap) and save yourself some LOC.

src/vivarium_public_health/disease/model.py
```diff
@@ -38,6 +38,21 @@ def __init__(self, cause, initial_state=None, get_data_functions=None, cause_typ
     def name(self):
         return f"disease_model.{self.cause}"
 
+    @property
+    def state_names(self):
+        return [s.name.split('.')[1] for s in self.states]
+
+    @property
+    def transition_names(self):
+        states = {s.name.split('.')[1]: s for s in self.states}
```
rmudambi, line 47: Why does this need to be a dict containing both the state name and the state itself? It seem like you are only using the value of the dict, so isn't it sufficient to iterate over `self.states`?

rmudambi (approved):

PR #231, opened 2021-06-29
Remove ad hoc type addition. Breaks docs build.

KjellSwedin:


src/vivarium_public_health/risks/distributions.py
```diff
@@ -74,7 +74,7 @@ def setup(self, builder):
                                                  creates_columns=[self._propensity],
                                                  requires_streams=[self._propensity])
 
-    def on_initialize_simulants(self, pop_data: 'SimulantData'):
+    def on_initialize_simulants(self, pop_data):
```
rmudambi, line 77: Is this the wrong type? Why does it break the docs build?

KjellSwedin: This is the only type addition in the file (and I added it). I decided just to pull it instead of adding extra imports. Seemed to me that adding typing could be a stand alone PR. James made the build process more rigorous and having the type information without the right imports generates a warning which causes the build to fail.

mattkappel (approved):

rmudambi (approved):

PR #244, opened 2021-11-30
refactor RiskEffect component to make it more easily sub-classed

rmudambi:


mattkappel (approved):

hussain-jafari (commented):
Are these changes just for readability?

src/vivarium_public_health/risks/effect.py
```diff
@@ -57,45 +65,70 @@ def __init__(self, risk: str, target: str):
...
+        self.configuration_defaults = self.get_configuration_defaults()
+
+    def __repr__(self):
+        return f"RiskEffect(risk={self.risk}, target={self.target})"
+
+    ##########################
+    # Initialization methods #
+    ##########################
+
+    def get_configuration_defaults(self) -> Dict[str, Dict]:
```
collijk, line 77: Either pass in risk and target as args (and make this static) or do this directly in the init.  I find stateful initialization subroutines to be a regular source of errors due to their dependence on sequencing of the init statements.

PR #248, opened 2021-12-22
Updated Black processing

mattkappel:
This PR includes updates Black processing for this repo as part of MIC-2556. That is, the effect of running `black` on the source top level directory.

This review is to determine if we like the formatting. Further work will be done if we do to integrate it and make efforts to preserve `git blame` behavior.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -21,17 +21,18 @@ from sphinx.ext.intersphinx import missing_reference
 
 
 import vivarium_public_health
+
 base_dir = Path(vivarium_public_health.__file__).parent
 
 about = {}
 with (base_dir / "__about__.py").open() as f:
     exec(f.read(), about)
 
-sys.path.insert(0, str(Path('..').resolve()))
+sys.path.insert(0, str(Path("..").resolve()))
 
 # -- Project information -----------------------------------------------------
 
-project = about['__title__']
+project = about["__title__"]
 copyright = f'2021, {about["__author__"]}'
 author = about["__author__"]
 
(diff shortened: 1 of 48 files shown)


albrja (approved):

src/vivarium_public_health/metrics/disability.py
```diff
@@ -37,67 +41,82 @@ class DisabilityObserver:
...
+
     configuration_defaults = {
-        'metrics': {
-            'disability': {
-                'by_age': False,
-                'by_year': False,
-                'by_sex': False,
-            }
-        }
+        "metrics": {"disability": {"by_age": False, "by_year": False, "by_sex": False,}}
```
rmudambi, line 46: I don't like how it condenses this into one line.

mattkappel: Agreed.

docs/source/conf.py
```diff
@@ -180,36 +188,44 @@
...
-                       'tables': ('https://www.pytables.org/', None),
-                       'numpy': ('https://numpy.org/doc/stable/', None),
-                       'vivarium': ('https://vivarium.readthedocs.io/en/latest/', None)}
+intersphinx_mapping = {
+    "python": ("https://docs.python.org/3.8", None),
+    "pandas": ("https://pandas.pydata.org/pandas-docs/stable/", None),
+    "tables": ("https://www.pytables.org/", None),
+    "numpy": ("https://numpy.org/doc/stable/", None),
+    "vivarium": ("https://vivarium.readthedocs.io/en/latest/", None),
+}
```
mattkappel, line 209: This is good reformatting

src/vivarium_public_health/disease/model.py
```diff
@@ -156,11 +205,15 @@ def get_state_weights(self, pop_index, prevalence_type):
         return state_names, weights_bins
 
     @staticmethod
-    def assign_initial_status_to_simulants(simulants_df, state_names, weights_bins, propensities):
-        simulants = simulants_df[['age', 'sex']].copy()
+    def assign_initial_status_to_simulants(
```
mattkappel, line 208: Not so great reformatting.

src/vivarium_public_health/metrics/disability.py
mattkappel: I wonder if it would have done what we're expecting if there was another key/value at the level of `"disability"`

PR #262, opened 2022-03-09
extend mortality component to account for unmodeled causes

rmudambi:
### Description

- *Category*: feature
- *JIRA issue*: [MIC-2914](https://jira.ihme.washington.edu/browse/MIC-2914)

Bring in the ability to account for csmr of unmodeled causes from CIFF
Reordered calls in `setup` to highlight the importance of `cause_specific_mortality_rate` and `mortality_rate`

### Testing
Ran CIFF model for a few time-steps with and without unmodeled causes.

### DIFF
--- a/src/vivarium_public_health/population/mortality.py
+++ b/src/vivarium_public_health/population/mortality.py
@@ -7,7 +7,7 @@ This module contains tools modeling all cause mortality and hooks for
 disease models to contribute cause-specific and excess mortality.
 
 """
-from typing import Callable
+from typing import Callable, Dict, List
 
 import pandas as pd
 from vivarium.framework.engine import Builder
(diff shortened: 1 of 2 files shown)


albrja (approved):

src/vivarium_public_health/population/mortality.py
```diff
@@ -83,13 +101,69 @@ def _get_mortality_rate(self, builder: Builder) -> Pipeline:
...
+
     # noinspection PyMethodMayBeStatic
     def _get_life_expectancy(self, builder: Builder) -> LookupTable:
         life_expectancy_data = builder.data.load(
             "population.theoretical_minimum_risk_life_expectancy"
         )
         return builder.lookup.build_table(life_expectancy_data, parameter_columns=["age"])
 
+    # noinspection PyMethodMayBeStatic
+    def _get_raw_unmodeled_csmr(self, builder: Builder) -> LookupTable:
```
collijk, line 119: Why not just put all this in the artifact under an "unmodeled_causes" key?

src/vivarium_public_health/population/mortality.py
```diff
@@ -83,13 +101,69 @@ def _get_mortality_rate(self, builder: Builder) -> Pipeline:
             requires_columns=["age", "sex"],
         )
 
+    # noinspection PyMethodMayBeStatic
+    def _get_all_cause_mortality_rate(self, builder: Builder) -> LookupTable:
```
collijk, line 105: Should type the returns as either `Union[LookupTable, Pipeline]` or `Callable[[pd.Index], Union[pd.Series, pd.DataFrame]]` to make the typing flexible in inheritance.

collijk (approved):
We should think about whether this structure of deferring all the logic of the setup method into a bunch of individual methods is the right long term move, and if so, how broadly it should be applied.  

As it stands, we've swung far in the direction of making the component easy to inherit and override in a mechanical sense.  However, it is extremely difficult to work out the actual construction of the mortality rate that's used to make decisions give the way it's laid out.  Ideally, we'd like both flexibility and clarity.  If we can't get there with more design on the component & configuration side, we should make some guidelines about how and when to do this sort of extreme structuring.

src/vivarium_public_health/population/mortality.py
rmudambi: I like that this makes it explicit in the config file which causes are included in the "unmodeled causes"

src/vivarium_public_health/population/mortality.py
rmudambi: I think vivarium should perhaps have a custom type that corresponds to something like this. It's complicated somewhat by the fact that those two type definitions you suggested aren't actually equivalent.

PR #264, opened 2022-03-25
preliminary port of observers; initial draft of stratifier and mortal…

rmudambi:
…ity observer

### Description

- *Category*: POC
- *JIRA issue*: [MIC-2915](https://jira.ihme.washington.edu/browse/MIC-2915)

POC of new design for Observers and ResultsStratifier. 
Can ignore all files except for `stratification.py`, `mortality.py` with regard to the POC. 
Other Observer files will be reworked in the same fashion as the MortalityObserver, so you can ignore them for now.
Removed several now unused methods in `utilities.py` and the respective test methods

I have some questions/ideas for further refinements in todo comments throughout the code.

### Testing
Ran toy version of CIFF
Should add unit tests
Will fix failing test suite

### DIFF
--- a/src/vivarium_public_health/metrics/mortality.py
+++ b/src/vivarium_public_health/metrics/mortality.py
@@ -7,13 +7,14 @@ This module contains tools for observing all-cause, cause-specific, and
 excess mortality in the simulation.
 
 """
-from vivarium_public_health.disease import DiseaseState, RiskAttributableDisease
-from vivarium_public_health.metrics.utilities import (
-    get_age_bins,
-    get_deaths,
-    get_person_time,
-    get_years_of_life_lost,
-)
+from typing import Dict, List
+
+import pandas as pd
+
+from vivarium.framework.engine import Builder, ConfigTree
+from vivarium.framework.population import PopulationView
+
+from vivarium_public_health.metrics.stratification import ResultsStratifier
 
 
 class MortalityObserver:
@@ -39,87 +40,97 @@ class MortalityObserver:
     """
 
     configuration_defaults = {
-        "metrics": {
-            "mortality": {
-                "by_age": False,
-                "by_year": False,
-                "by_sex": False,
+        "include": [ResultsStratifier.DEATH_YEAR, ResultsStratifier.CAUSE_OF_DEATH],
+        "exclude": [ResultsStratifier.YEAR],
+    }
+
+    def __init__(self):
+        self.configuration_defaults = self._get_configuration_defaults()
+
+        self.metrics_pipeline_name = "metrics"
+        self.tmrle_key = "population.theoretical_minimum_risk_life_expectancy"
+
+    ##########################
+    # Initialization methods #
+    ##########################
+
+    # noinspection PyMethodMayBeStatic
+    def _get_configuration_defaults(self) -> Dict[str, Dict]:
+        return {
+            "metrics": {
+                "mortality": MortalityObserver.configuration_defaults
             }
         }
-    }
+
+    ##############
+    # Properties #
+    ##############
 
     @property
     def name(self):
         return "mortality_observer"
 
-    def setup(self, builder):
-        self.config = builder.configuration.metrics.mortality
-        self.clock = builder.time.clock()
-        self.step_size = builder.time.step_size()
-        self.start_time = self.clock()
-        self.initial_pop_entrance_time = self.start_time - self.step_size()
-        self.age_bins = get_age_bins(builder)
-        diseases = builder.components.get_components_by_type(
-            (DiseaseState, RiskAttributableDisease)
-        )
-        self.causes = [c.state_id for c in diseases] + ["other_causes"]
+    #################
+    # Setup methods #
+    #################
 
-        life_expectancy_data = builder.data.load(
-            "population.theoretical_minimum_risk_life_expectancy"
-        )
-        self.life_expectancy = builder.lookup.build_table(
-            life_expectancy_data, key_columns=[], parameter_columns=["age"]
-        )
+    # noinspection PyAttributeOutsideInit
+    def setup(self, builder: Builder):
+        self.config = self._get_stratification_configuration(builder)
+        self.stratifier = self._get_stratifier(builder)
+        self.population_view = self._get_population_view(builder)
 
+        self.register_metrics_modifier(builder)
+
+    # noinspection PyMethodMayBeStatic
+    def _get_default_stratifications(self, builder: Builder) -> List[str]:
+        return builder.configuration.metrics.default
+
+    # noinspection PyMethodMayBeStatic
+    def _get_stratification_configuration(self, builder: Builder) -> ConfigTree:
+        return builder.configuration.metrics.mortality
+
+    # noinspection PyMethodMayBeStatic
+    def _get_stratifier(self, builder: Builder) -> ResultsStratifier:
+        return builder.components.get_component(ResultsStratifier.NAME)
+
+    # noinspection PyMethodMayBeStatic
+    def _get_population_view(self, builder: Builder) -> PopulationView:
         columns_required = [
             "tracked",
             "alive",
-            "entrance_time",
-            "exit_time",
-            "cause_of_death",
             "years_of_life_lost",
             "age",
         ]
-        if self.config.by_sex:
-            columns_required += ["sex"]
-        self.population_view = builder.population.get_view(columns_required)
+        return builder.population.get_view(columns_required)
 
-        builder.value.register_value_modifier("metrics", self.metrics)
+    def register_metrics_modifier(self, builder: Builder) -> None:
+        builder.value.register_value_modifier(
+            self.metrics_pipeline_name,
+            modifier=self.metrics,
+            requires_columns=["age", "exit_time", "alive"],
+        )
 
-    def metrics(self, index, metrics):
+    ########################
+    # Event-driven methods #
+    ########################
+
+    def metrics(self, index: pd.Index, metrics: Dict) -> Dict:
         pop = self.population_view.get(index)
-        pop.loc[pop.exit_time.isnull(), "exit_time"] = self.clock()
 
-        person_time = get_person_time(
-            pop, self.config.to_dict(), self.start_time, self.clock(), self.age_bins
-        )
-        deaths = get_deaths(
-            pop,
-            self.config.to_dict(),
-            self.start_time,
-            self.clock(),
-            self.age_bins,
-            self.causes,
-        )
-        ylls = get_years_of_life_lost(
-            pop,
-            self.config.to_dict(),
-            self.start_time,
-            self.clock(),
-            self.age_bins,
-            self.life_expectancy,
-            self.causes,
-        )
-
-        metrics.update(person_time)
-        metrics.update(deaths)
-        metrics.update(ylls)
+        groups = self.stratifier.group(pop, set(self.config.include), set(self.config.exclude))
+        for label, pop_in_group in groups:
+            new_observations = {
+                f"death_{label}": pop_in_group.size,
+                f"ylls_{label}": pop_in_group["years_of_life_lost"].sum()
+            }
+            metrics.update(new_observations)
 
         the_living = pop[(pop.alive == "alive") & pop.tracked]
         the_dead = pop[pop.alive == "dead"]
-        metrics["years_of_life_lost"] = self.life_expectancy(the_dead.index).sum()
-        metrics["total_population_living"] = len(the_living)
-        metrics["total_population_dead"] = len(the_dead)
+        metrics["years_of_life_lost"] = the_dead["years_of_life_lost"].sum()
+        metrics["total_population_living"] = the_living.size
+        metrics["total_population_dead"] = the_dead.size
 
         return metrics
 
--- a/src/vivarium_public_health/metrics/mortality.py
+++ b/src/vivarium_public_health/metrics/mortality.py
@@ -143,10 +143,15 @@ class MortalityObserver:
         pop = self.population_view.get(event.index)
         pop_died = pop[(pop["alive"] == "dead") & (pop["exit_time"] > event.time - self.time_step)]
 
-        groups = self.stratifier.group(pop_died, set(self.config.include), set(self.config.exclude))
-        for label, pop_in_group in groups:
+        groups = self.stratifier.group(
+            pop_died.index, set(self.config.include), set(self.config.exclude)
+        )
+        for label, group_index in groups:
             for cause in self.causes_of_death:
-                pop_died_of_cause = pop_in_group.query(f"cause_of_death == '{cause}'")
+                pop_died_of_cause = (
+                    pop_died.loc[group_index, :]
+                    .query(f"cause_of_death == '{cause}'")
+                )
                 new_observations = {
                     f"death_due_to_{cause}_{label}": pop_died_of_cause.size,
                     f"ylls_due_to_{cause}_{label}": pop_died_of_cause["years_of_life_lost"].sum()
--- a/src/vivarium_public_health/metrics/mortality.py
+++ b/src/vivarium_public_health/metrics/mortality.py
@@ -7,13 +7,19 @@ This module contains tools for observing all-cause, cause-specific, and
 excess mortality in the simulation.
 
 """
-from typing import Dict
+from collections import Counter
+from typing import Dict, Set
+
 import pandas as pd
 
 from vivarium.framework.engine import Builder, ConfigTree
+from vivarium.framework.event import Event
 from vivarium.framework.population import PopulationView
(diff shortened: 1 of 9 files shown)


albrja (approved):

rmudambi:
Addressed initial PR comments regarding stratification of `MortalityObserver` by cause and year. Changed observer to check for deaths on each timestep so it can use year in the same way as other observers. Also changed to pass an index rather than a dataframe to `group()`

I am now enforcing the order of stratifications to be the same as the order in which they are registered.

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -7,121 +7,173 @@
...
-            "mortality": {
-                "by_age": False,
-                "by_year": False,
-                "by_sex": False,
+        "exclude": [],
+        "include": [],
+    }
+
+    def __init__(self):
+        self.configuration_defaults = self._get_configuration_defaults()
```
collijk, line 56: You can drop this configuration defaults stuff.  Mortality is not a parameterizable component, so subclasses can just overwrite the configuration defaults.

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -7,121 +7,173 @@
...
 
     """
 
     configuration_defaults = {
-        "metrics": {
-            "mortality": {
-                "by_age": False,
-                "by_year": False,
-                "by_sex": False,
+        "exclude": [],
```
collijk, line 51: Must be 
```
configuration_defaults = {
    "observers": {
        "mortality": {
            "include": [],
            "exclude": [],
        }
    }
```

All non-parameterizable components are explicit about their configuration name space.  We can discuss doing something different, but we'd want to make that change uniformly.

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -7,121 +7,173 @@
...
-        metrics.update(person_time)
-        metrics.update(deaths)
-        metrics.update(ylls)
+    ########################
+    # Event-driven methods #
+    ########################
+
+    def on_collect_metrics(self, event: Event) -> None:
+        pop = self.population_view.get(event.index)
+        pop_died = pop[(pop["alive"] == "dead") & (pop["exit_time"] > event.time - self.step_size)]
```
collijk, line 148: I think doing exact equality `pop["exit_time"] == event.time` is fine here.  I can't remember right now why we said something different.

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -7,121 +7,173 @@
...
 class MortalityObserver:
     """An observer for cause-specific deaths, ylls, and total person time.
 
-    By default, this counts cause-specific deaths, years of life lost, and
-    total person time over the full course of the simulation. It can be
-    configured to bin these measures into age groups, sexes, and years
-    by setting the ``by_age``, ``by_sex``, and ``by_year`` flags, respectively.
+    By default, this counts cause-specific deaths and years of life lost over
+    the full course of the simulation. It can be configured to add or remove
+    stratification groups to the default groups defined by a ResultsStratifier.
```
collijk, line 30: Use `:class:ResultsStratifier` so the docs link to the the stratifier class when built.

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -7,121 +7,173 @@
...
-            metrics:
+            observers:
                 mortality:
-                    by_age: True
-                    by_year: False
-                    by_sex: True
+                    exclude:
+                        - "year"
+                    include:
+                        - "death_year"
```
collijk, line 43: I think you'll want to write a document about the default + include/exclude logic used to determine what is stratified by what and then link to it in these observer docstrings.  Otherwise this code example is not possible to interpret cause there's no context for how it's used.

collijk: Also, `"death_year"` should have gone away to be replaced by `"year"` and `"cause_of_death"` is no longer a stratification level.

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -7,121 +7,173 @@
...
+
+        self._register_collect_metrics_listener(builder)
+        self._register_metrics_modifier(builder)
+
+    # noinspection PyMethodMayBeStatic
+    def _get_stratification_configuration(self, builder: Builder) -> ConfigTree:
+        return builder.configuration.observers.mortality
+
+    # noinspection PyMethodMayBeStatic
+    def _get_step_size(self, builder: Builder) -> Timedelta:
```
collijk, line 106: Don't need this

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -7,121 +7,173 @@
...
-        self.initial_pop_entrance_time = self.start_time - self.step_size()
-        self.age_bins = get_age_bins(builder)
+    #################
+    # Setup methods #
+    #################
+
+    # noinspection PyAttributeOutsideInit
+    def setup(self, builder: Builder):
+        self.config = self._get_stratification_configuration(builder)
+        self.step_size = self._get_step_size(builder)
```
collijk, line 91: Don't need this

src/vivarium_public_health/metrics/mortality.py
collijk: If you do need the `step_size` it's available as an attribute on the `event`.

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -7,121 +7,173 @@
...
+    ########################
+    # Event-driven methods #
+    ########################
+
+    def on_collect_metrics(self, event: Event) -> None:
+        pop = self.population_view.get(event.index)
+        pop_died = pop[(pop["alive"] == "dead") & (pop["exit_time"] > event.time - self.step_size)]
+
+        groups = self.stratifier.group(
+            pop_died.index, set(self.config.include), set(self.config.exclude)
```
collijk, line 151: Allow `group` to take lists as args and coerce to sets on the stratifier side.  Generally if you can make the code in the utility function a little uglier in one place to allow client code to be cleaner in a lot of places, you should do it.

collijk: Idea for later (unless it's super easy and you agree for now): 

It may be nice to allow the user to provide an `as_mask` arg to `ResultsStratifier.group` that returns a boolean mask on the provided index to indicate the group.  This would support

```
groups = self.stratifier.group(
    pop_died.index, set(self.config.include), set(self.config.exclude)
)
for label, group_index in groups:
    for cause in self.causes_of_death:
        pop_died_of_cause = (
            pop_died.loc[group_index, :]
            .query(f"cause_of_death == '{cause}'")
        )
```
and 

```
groups = self.stratifier.group(
    pop_died.index, set(self.config.include), set(self.config.exclude), as_mask=True,
)
for label, group_mask in groups:
    for cause in self.causes_of_death:
        cause_mask = pop_died['cause_of_death'] == cause
        pop_died_of_cause = pop_died.loc[group_mask & cause_mask]
```
The mask syntax is going to allow for more flexibility (e.g. that doesn't rely on query syntax).  

It's also faster.  Simple example I have handy:

image

collijk: Speedup is less profound if things are strings, so caveat: 
image

src/vivarium_public_health/metrics/risk.py
```diff
@@ -125,22 +124,6 @@ def on_time_step_prepare(self, event: Event):
             )
             self.person_time.update(state_person_time_this_step)
 
-        if self._should_sample(event.time):
```
collijk, line 128: :+1:

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -0,0 +1,353 @@
+from dataclasses import dataclass
```
collijk, line 1: I think this will break 3.6 support without a future import, I'm pretty sure.

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -0,0 +1,353 @@
...
+    """Centralized component for handling results stratification.
+
+    This should be used as a subcomponent for observers.  The observers
+    can then ask this component for population subgroups and labels during
+    results production and have this component manage adjustments to the
+    final column labels for the subgroups.
+
+    """
+
+    NAME = "results_stratifier"
```
collijk, line 65: Not sure why you want this

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -0,0 +1,353 @@
...
+
+    configuration_defaults = {
+        "default": [],
```
collijk, line 68: Again, non-parameterizable component, so namespace should be complete here

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    def __init__(self):
+        self.configuration_defaults = self._get_configuration_defaults()
```
collijk, line 72: And not do this

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -0,0 +1,353 @@
...
+
+        self.metrics_pipeline_name = "metrics"
+        self.tmrle_key = "population.theoretical_minimum_risk_life_expectancy"
+
+    ##########################
+    # Initialization methods #
+    ##########################
+
+    # noinspection PyMethodMayBeStatic
+    def _get_configuration_defaults(self) -> Dict[str, Dict]:
```
collijk, line 82: or this

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -0,0 +1,353 @@
...
+            }
+        }
+
+    ##############
+    # Properties #
+    ##############
+
+    @property
+    def name(self) -> str:
+        return ResultsStratifier.NAME
```
collijk, line 95: So, even if you have a reason for the class variable, you're now in a situation where client code, which holds on to the results stratifier, can access both the `.name` attribute and the `.NAME` attribute, which are hopefully the same.  Both of which can be modified by subclasses though.  

I'd make the class constant `name` and just not have an instance property for what I think is your intended use case (looking up the stratifier in other observers).

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -0,0 +1,353 @@
...
+
+    #################
+    # Setup methods #
+    #################
+
+    # noinspection PyAttributeOutsideInit
+    def setup(self, builder: Builder) -> None:
+        """Perform this component's setup."""
+        self.clock = None
```
collijk, line 104: Always get a clock

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -0,0 +1,353 @@
...
+        )
+        self.stratification_levels[stratification_level.name] = stratification_level
+
+        for source in stratification_level.sources:
+            if source.type == SourceType.PIPELINE:
+                self.pipelines[source.name] = builder.value.get_value(source.name)
+            elif source.type == SourceType.COLUMN:
+                self.columns_required.append(source.name)
+            elif source.type == SourceType.CLOCK:
+                self.clock = self._get_clock(builder)
```
collijk, line 295: Yeah, definitely don't want to assign object attributes way down here.

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -0,0 +1,353 @@
...
+        self.age_bins = self._get_age_bins(builder)
+
+        self.register_stratifications(builder)
+        self.population_view = self._get_population_view(builder)
+
+        self._register_timestep_prepare_listener(builder)
+        self._register_simulation_end_listener(builder)
+
+    # noinspection PyMethodMayBeStatic
+    def _get_clock(self, builder: Builder) -> Callable[[], Time]:
```
collijk, line 121: Just for thought on the structure of components and their inheritibility:

The layer of indirection that you make to access builder services is a signal about what is available and potentially _intended_ to be modified during inheritance.  The thing to keep in mind then, is that we have several kinds of things (e.g. the timekeeping system) that are meant to be modified by configuration and not by inheritance.

src/vivarium_public_health/metrics/mortality.py
rmudambi: Good call on documentation. death_year I fixed in a later PR, but I'll move it into this one.

src/vivarium_public_health/metrics/mortality.py
rmudambi: Works for me. I'm actually using masks in the other observers now that I look, so this will make things simpler there as well

src/vivarium_public_health/metrics/stratification.py
rmudambi: You are correct. There is a backport for `dataclasses` which appears to be legit. Users shouldn't be using 3.6 ideally, but this worked with the backport when I tested it.

src/vivarium_public_health/metrics/stratification.py
rmudambi: Observer components will get the ResultsStratifier by name. This is the constant to be used for that purpose

src/vivarium_public_health/metrics/stratification.py
rmudambi: Good point. I'll consider what things can be changed in light of that observation.

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -7,121 +7,152 @@
...
     .. code-block:: yaml
 
         configuration:
-            metrics:
+            observers:
                 mortality:
-                    by_age: True
-                    by_year: False
-                    by_sex: True
+                    exclude:
```
beatrixkh, line 39: instead of referring on line 29 to the default groups defined by ResultsStratifier, it might be nice to note what the defaults are here, so it's easy to know what might need to be `exclude`d?

or perhaps this is something that we do so often that that's unnecessary? i'm not sure

beatrixkh: oh --- hm, reading the `ResultsStratifier` i think you mean for the defaults to be customizable in a given project. If that's correct i retract the above comment.

rmudambi: yes exactly. the defaults are configurable in the models themselves!

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -0,0 +1,319 @@
...
+
+    configuration_defaults = {
+        "observers": {
+            "default": [],
+        }
+    }
+
+    def __init__(self):
+        self.metrics_pipeline_name = "metrics"
+        self.tmrle_key = "population.theoretical_minimum_risk_life_expectancy"
```
beatrixkh, line 75: sp: `tmrel`

mattkappel: No, this is right. TMREL is different.

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -7,121 +7,152 @@
...
-                "by_sex": False,
+                "exclude": [],
+                "include": [],
             }
         }
     }
 
+    def __init__(self):
+        self.metrics_pipeline_name = "metrics"
+        self.tmrle_key = "population.theoretical_minimum_risk_life_expectancy"
```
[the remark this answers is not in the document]
mattkappel: No, this is right. TMREL is different.

mattkappel (approved):

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -0,0 +1,319 @@
...
+@dataclass
+class Source:
+    # todo add docstrings
+    name: str
+    type: SourceType
+
+
+@dataclass
+class StratificationLevel:
+    # todo add docstrings
```
beatrixkh, line 27: It would be nice to have the docstrings in this PR

src/vivarium_public_health/metrics/mortality.py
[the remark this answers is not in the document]
rmudambi: this is actually correct. this isn't the TMREL (theoretical minimum risk exposure level), but the TMRLE (theoretical minimum risk life expectancy)

src/vivarium_public_health/metrics/stratification.py
beatrixkh: oh! oops, thank you!

rmudambi: see above

src/vivarium_public_health/metrics/stratification.py
rmudambi: I'll add them in a separate PR, if that's alright.

PR #266, opened 2022-03-31
filter causes without excess mortality or disability from their respe…

rmudambi:
…ctive observers

### Description

- *Category*: feature
- *JIRA issue*: [MIC-2982](https://jira.ihme.washington.edu/browse/MIC-2982)

Filter out causes that have no excess mortality from the MortalityObserver
Filter out causes that have no disability from the DisabilityObserver

This has to happen in post_setup, because all DiseaseState components must be set-up prior to the observers

### Testing
Ran toy version of CIFF

### DIFF
--- a/src/vivarium_public_health/metrics/disability.py
+++ b/src/vivarium_public_health/metrics/disability.py
@@ -113,9 +113,11 @@ class DisabilityObserver:
 
     # noinspection PyMethodMayBeStatic
     def _get_cause_components(
-            self, builder: Builder
+        self, builder: Builder
     ) -> List[Union[DiseaseState, RiskAttributableDisease]]:
-        return builder.components.get_components_by_type((DiseaseState, RiskAttributableDisease))
+        return builder.components.get_components_by_type(
+            (DiseaseState, RiskAttributableDisease)
+        )
 
     # noinspection PyMethodMayBeStatic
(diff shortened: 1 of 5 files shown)


albrja (approved):

src/vivarium_public_health/metrics/disability.py
```diff
@@ -150,6 +152,12 @@ def _register_metrics_modifier(self, builder: Builder) -> None:
     # Event-driven methods #
     ########################
 
+    # noinspection PyUnusedLocal
+    def on_post_setup(self, event: Event) -> None:
```
beatrixkh, line 156: for my own understanding: is this post-setup step new / is its sole purpose to check if the cause is applicable?

beatrixkh (approved):

mattkappel (approved):

src/vivarium_public_health/metrics/disability.py
rmudambi: This is because we don't have guarantees on the order in which components have their setup methods run. The disability and mortality components need all components that potentially have excess disability weight or excess mortality to have run their setup so that the `has_disability_weight` or `has_excess_mortality` fields have been created on them.

hussain-jafari (approved):

PR #270, opened 2022-04-12
throw an error if a mapper produces an invalid category

rmudambi:
### Description

- *Category*: bugfix
- *JIRA issue*: [MIC-3016](https://jira.ihme.washington.edu/browse/MIC-3016)

Wrap mappers so that that they throw an error if they produce an illegal value
Add docstrings.

### Testing
Ran test suite.
Ran toy CIFF model.

### DIFF
--- a/src/vivarium_public_health/metrics/stratification.py
+++ b/src/vivarium_public_health/metrics/stratification.py
@@ -26,14 +26,36 @@ class SourceType(Enum):
 
 @dataclass
 class Source:
-    # todo add docstrings
+    """
+    A source of information about simulants used to determine which category
+    they belong to for a given stratification level. The source name should be
+    the name of the column or pipeline being used as the source. If the source
+    is of type clock, the name should be something descriptive and unique.
+    """
     name: str
     type: SourceType
 
 
 @dataclass
 class StratificationLevel:
-    # todo add docstrings
+    """
+    A level of stratification. Each StratificationLevel represents a set of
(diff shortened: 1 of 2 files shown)


albrja (approved):

src/vivarium_public_health/metrics/stratification.py
```diff
@@ -26,50 +26,94 @@ class SourceType(Enum):
...
     name: str
     sources: List[Source]
     categories: Set[str]
     mapper: Callable[[pd.Series], str] = None
     current_categories_getter: Callable[[], Set[str]] = None
 
     def __post_init__(self):
-        self.mapper = self.mapper if self.mapper else self._default_mapper
+        self._set_mapper()
+        self._set_current_categories_getter()
```
hussain-jafari, line 67: Do we use self._set_current_categories_getter() instead of using what's in the body of the function to define self.current_categories_getter for consistency between these components? I was thinking we were setting the mapper so that we could wrap it around the category validator but I don't see an equivalent reason to do the same for the current categories getter.

hussain-jafari (approved):

src/vivarium_public_health/metrics/stratification.py
rmudambi: There isn't a fundamental reason for why I changed how I'm setting the current categories getter. I just changed it like this to be consistent with the `self._set_mapper()` function

mattkappel (approved):

PR #277, opened 2022-04-26
implement LBWSG risk effect component

rmudambi:
### Description

- *Category*: feature
- *JIRA issue*: [MIC-3043](https://jira.ihme.washington.edu/browse/MIC-3043)

Created a LBWSG Risk Effect component
Refactored the AXES constant
Created a utility to convert a string to snake case

### Testing
Ran CIFF model

### DIFF
--- a/src/vivarium_public_health/risks/implementations/low_birth_weight_and_short_gestation.py
+++ b/src/vivarium_public_health/risks/implementations/low_birth_weight_and_short_gestation.py
@@ -117,6 +117,11 @@ class LBWSGDistribution(PolytomousDistribution):
         and either a categorical propensity or a categorical exposure and
         returns continuous exposures for that axis.
 
+        If categorical propensity is provided rather than exposure, this
+        function requires access to the low birth weight and short gestation
+        categorical exposure parameters pipeline
+        ("risk_factor.low_birth_weight_and_short_gestation.exposure_parameters").
+
         :param axis:
         :param propensity:
         :param categorical_propensity:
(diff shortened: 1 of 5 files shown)


src/vivarium_public_health/risks/implementations/low_birth_weight_and_short_gestation.py
```diff
@@ -265,3 +267,178 @@ def _get_current_exposure(self, index: pd.Index) -> pd.DataFrame:
             f" refer directly one of the exposure columns. During simulant initialization the birth"
             f" exposure pipelines should be used instead."
         )
+
+
+class LBWSGRiskEffect(RiskEffect):
+
+    TMREL_BIRTH_WEIGHT_INTERVAL: pd.Interval = pd.Interval(3500.0, 4500.0)
```
beatrixkh, line 274: is this not stored in the data just because it will never change?

beatrixkh (approved):

src/vivarium_public_health/risks/implementations/low_birth_weight_and_short_gestation.py
mattkappel: Good question, also, not being familiar, what is the significance of these particular values? Maybe a comment or constant assignment is in order?

rmudambi: This has to do with interpolation. The categorical risk effects for all categories in the area will have RR == 1, but when we perform an interpolation, that would no longer be true so that the surface would be continuous. This introduces a discontinuity at the boundary of TMREL which is not reflected in the interpolator itself. You can see how this is used in `get_relative_risk_for_age_group` called in `on_initialize_simulants`.

mattkappel (approved):

src/vivarium_public_health/risks/implementations/low_birth_weight_and_short_gestation.py
mattkappel: ^ should that detail be noted in the code?

PR #294, opened 2022-12-23
Update ci, readme, and setup

stevebachmeier:
### Description
- *Category*: other
- *JIRA issue*: [MIC-3690](https://jira.ihme.washington.edu/browse/MIC-3690)

This addresses a few issues with the github actions:
* Updates the python builds 3.7-3.10 (remove 3.6 and add 3.9-10)
* Update actions to stop future warnings (The `set-output` command
    is deprecated and will be disabled soon. Please upgrade to using
    Environment Files. For more information see: 
    https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)
* Update CODEOWNERS
* Update README install 3.10
* Update setup to use min 3.7

### Testing
All tests passed and warnings are gone

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 # -----------------------------------------------------------------------------
 #   - invoked on push, pull_request, or manual trigger
-#   - test under 3 versions of python
+#   - test under at least 3 versions of python
 # -----------------------------------------------------------------------------
 name: build
 on: [push, pull_request, workflow_dispatch]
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -10,14 +10,14 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
     defaults:
       run:
-        shell: bash -l {0}
+        shell: bash -le {0}
```
hussain-jafari, line 16: What does this do?

setup.py
```diff
@@ -22,7 +22,7 @@
         "scipy",
         "tables",
         "risk_distributions>=2.0.10",
-        "dataclasses; python_version < '3.7.0'",
+        "dataclasses",
```
rmudambi, line 25: This is part of core python. We shouldn't need to import it explicitly

rmudambi (approved):

.github/workflows/build.yml
stevebachmeier: Zeb found a long-standing bug where we'd inadvertently removed the -e flag which immediately stops the script upon a non-zero exit code. (https://github.com/ihmeuw/vivarium_census_prl_synth_pop/pull/108)

PR #315, opened 2023-05-25
Handle FutureWarning: Calling float on a single element Series

mattkappel:
### Description
- *Category*: bugfix
- *JIRA issue*: [MIC-3599](https://jira.ihme.washington.edu/browse/MIC-3599)

### Changes and notes
Follows suggestion from FutureWarning to get the single Series item via .iloc[0]

### Testing
Tested with the Maternal IV Iron simulation. When the FutureWarning ignore filters are removed from vivarium, previously emitted FutureWarnings no longer appear with this change.

### DIFF
--- a/src/vivarium_public_health/population/data_transformations.py
+++ b/src/vivarium_public_health/population/data_transformations.py
@@ -108,13 +108,13 @@ def rescale_binned_proportions(
         values are rescaled to reflect their smaller representation.
     """
     col_order = pop_data.columns.copy()
-    if age_start > pop_data.age_end.max():
+    if age_start > pop_data["age_end"].max():
         raise ValueError(
             "Provided population data is insufficient to model the requested age range."
         )
 
-    age_start = max(pop_data.age_start.min(), age_start)
-    age_end = min(pop_data.age_end.max(), age_end) - 1e-8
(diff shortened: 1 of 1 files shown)


src/vivarium_public_health/population/data_transformations.py
```diff
@@ -123,12 +123,11 @@ def rescale_binned_proportions(
         "value",
     ]
     for _, sub_pop in pop_data.groupby(["sex", "location"]):
-
         min_bin = sub_pop[(sub_pop.age_start <= age_start) & (age_start < sub_pop.age_end)]
-        padding_bin = sub_pop[sub_pop.age_end == float(min_bin.age_start)]
+        padding_bin = sub_pop[sub_pop.age_end == float(min_bin.age_start.iloc[0])]
```
rmudambi, line 127: Nit: We generally prefer using bracket notation rather than dot notation. This seems like a good excuse to clean that up in this function.

rmudambi (approved):

src/vivarium_public_health/population/data_transformations.py
mattkappel: Good call. I'll update.

hussain-jafari (approved):

stevebachmeier (approved):

PR #321, opened 2023-06-15
change observers to stratification in config example

hussain-jafari:
change observers to stratification in config example


mattkappel (approved):

mattkappel:
Thanks for catching this!

PR #326, opened 2023-07-18
update bmi distribution xmin and xmax

hussain-jafari:
### Description
- *Category*: test
- *JIRA issue*: [MIC-4242](https://jira.ihme.washington.edu/browse/MIC-4242)

### Changes and notes
Update distributions in BMI ensemble to have an xmin and xmax of 10 and 50 respectively. 

NOTE: this will not be merged 

### Testing
Put a breakpoint in the setup of the distribution and checked the distribution parameters before the ensemble was setup.

### DIFF
--- a/src/vivarium_public_health/risks/distributions.py
+++ b/src/vivarium_public_health/risks/distributions.py
@@ -39,6 +39,13 @@ class SimulationDistribution:
     def setup(self, builder):
         distribution_data = get_distribution_data(builder, self.risk)
         self.implementation = get_distribution(self.risk, **distribution_data)
+        if self.risk == 'risk_factor.high_body_mass_index_in_adults':
+            for dist in self.implementation._parameters:
+                dist_params = self.implementation._parameters[dist]
+                dist_params.loc[dist_params['x_min'] != 0, 'x_min'] = 10
+                dist_params.loc[dist_params['x_max'] != 0, 'x_max'] = 50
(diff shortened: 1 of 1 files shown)


src/vivarium_public_health/risks/distributions.py
```diff
@@ -39,6 +39,13 @@ def name(self):
     def setup(self, builder):
         distribution_data = get_distribution_data(builder, self.risk)
         self.implementation = get_distribution(self.risk, **distribution_data)
+        if self.risk == "risk_factor.high_body_mass_index_in_adults":
+            for dist in self.implementation._parameters:
+                dist_params = self.implementation._parameters[dist]
+                dist_params.loc[dist_params["x_min"] != 0, "x_min"] = 10
+                dist_params.loc[dist_params["x_max"] != 0, "x_max"] = 50
```
stevebachmeier, line 46: How is this working? I would've thought you'd need to subset to values that are outside your thresholds, e.g. `dist_params.loc[dist_params["x_max"] > 50, "x_max"] = 50`

hussain-jafari: There were some values were the xmin and xmax were 0 (we didn't have data for infants for example), which I wanted to keep 0 to prevent any unintended behavior. I also verified that we didn't have any values greater than x_max or less than x_min, so we didn't need to check that.

PR #327, opened 2023-07-20
Make Mortality Observer act on Collect Metrics

patricktnast:
### Description

- *Category*:  bugfix
- *JIRA issue*: [MIC-4258](https://jira.ihme.washington.edu/browse/MIC-4258)

### Changes and notes
 

The mortality observer `on_collect_metrics` listener is actually registered to `timestep`. Normally this works fine, since the standard mortality handler has a `priority` of `0`, but if you change this, deaths stop being recorded by the observer because the event times are mismatched. It seems that it should be registered to 'collect_metrics' instead.
### Testing

This was causing my sim to not register deaths on the right event time, and now it works the way I expect, and presume was intended. It doesn't look like there are unit tests for the mortality component, so I didn't write a new one, but could if requested.

### DIFF
--- a/src/vivarium_public_health/metrics/mortality.py
+++ b/src/vivarium_public_health/metrics/mortality.py
@@ -114,7 +114,7 @@ class MortalityObserver:
         builder.event.register_listener("post_setup", self.on_post_setup)
 
     def _register_collect_metrics_listener(self, builder: Builder) -> None:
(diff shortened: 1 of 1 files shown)


albrja (approved):

mattkappel:
Good catch!

mattkappel (approved):

hussain-jafari (approved):

PR #337, opened 2023-08-09
make population components subclass Component from vivarium

rmudambi:
### Description

- *Category*: refactor/POC
- *JIRA issue*: [MIC-4329](https://jira.ihme.washington.edu/browse/MIC-4329)

### Changes and notes
Make `BasePopulation` and `AgeOutSimulants` inherit from `Component`

### Testing
Tests pass locally and I can run a simulation.

### DIFF
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -11,9 +11,10 @@ from typing import Callable, Dict, Iterable, List
 
 import numpy as np
 import pandas as pd
+from vivarium import VivariumComponent
 from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
-from vivarium.framework.population import PopulationView, SimulantData
+from vivarium.framework.population import SimulantData
 from vivarium.framework.randomness import RandomnessStream
 
 from vivarium_public_health import utilities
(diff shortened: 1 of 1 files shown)


hussain-jafari (approved):

src/vivarium_public_health/population/base_population.py
```diff
@@ -175,20 +176,41 @@ def get_demographic_proportions_for_creation_time(
...
     # noinspection PyAttributeOutsideInit
     def setup(self, builder: Builder) -> None:
-        if builder.configuration.population.exit_age is None:
-            return
-        self.config = builder.configuration.population
-        self.population_view = builder.population.get_view(["age", "exit_time", "tracked"])
-        builder.event.register_listener("time_step__cleanup", self.on_time_step_cleanup)
+        if builder.configuration.population.exit_age is not None:
+            self.config = builder.configuration.population
+            self._columns_required.extend(["age", "exit_time", "tracked"])
```
stevebachmeier, line 210: Is this behaving as intended? This extends the required cols whereas it used to re-define them.

stevebachmeier: oh wait, nm they're always just empty at this point so effectively the same thing

stevebachmeier (approved):

src/vivarium_public_health/population/base_population.py
rmudambi: It used not to define `self.population_view` at all, since there was no need for one. Now we are setting it to be None, rather than not existing at all. 

What I'm doing here is adding these columns to be required if there is an exit age (which will make it so that a population view is created with those columns when we get to the `super` call below.

PR #344, opened 2023-09-07
refactor disease model components accommodate inheriting from Component

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: [MIC-4473](https://jira.ihme.washington.edu/browse/MIC-4473)

### Changes and notes
- Simplifies `AgeOutSimulants`' time-step cleanup method - this is changed in this PR because the test started failing
- Refactors `DiseaseModel` and the various state and transition classes to accommodate the fact that their super-classes now inherit from `Component`. 
- Adds some typing where it wasn't increasing the scope of the PR too much
- Replaced previous `state.add_transition()` functions with `state.add_rate_transition()`, `state.add_proportion_transition()`, and `state.add_dwell_time_transition` which I think are clearer, and bring us in line with the new signature of `add_transition` in `State`. This also means we no longer have sub-classes having a method with the same name, but incompatible arguments as their super-class.

### Testing
All unit tests pass.

### DIFF
--- a/src/vivarium_public_health/disease/model.py
+++ b/src/vivarium_public_health/disease/model.py
@@ -8,14 +8,17 @@ function is to provide coordination across a set of disease states and
 transitions at simulation initialization and during transitions.
 
 """
-from typing import List
+from typing import Callable, Dict, List, Optional
 
 import numpy as np
 import pandas as pd
 from vivarium.exceptions import VivariumError
+from vivarium.framework.engine import Builder
+from vivarium.framework.event import Event
+from vivarium.framework.population import SimulantData
 from vivarium.framework.state_machine import Machine
 
-from vivarium_public_health.disease.state import SusceptibleState
+from vivarium_public_health.disease.state import BaseDiseaseState, SusceptibleState
 from vivarium_public_health.disease.transition import TransitionString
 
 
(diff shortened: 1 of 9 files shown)


src/vivarium_public_health/disease/special_disease.py
```diff
@@ -322,6 +346,7 @@ def continuous_filter(index):
 
         return filter_function
 
+    # todo: reorder these methods to make it easier to understand what they are each doing
```
patricktnast, line 349: I think now would be a good time to address this todo.

src/vivarium_public_health/disease/state.py
```diff
@@ -102,110 +120,124 @@ def _transition_side_effect(self, index, event_time):
...
+        """
+        if "proportion" not in get_data_functions:
+            raise ValueError("You must supply a proportion function.")
+
+        transition = ProportionTransition(self, output, get_data_functions, **kwargs)
+        self.add_transition(transition)
+        return transition
+
+
+# TODO is there a better name for the union of susceptible and recovered?
```
patricktnast, line 182: ChatIHME suggests "non-diseased" is an epidemiological term of art for this, and I think it is correct based on a quick follow-up

patricktnast (approved):

src/vivarium_public_health/disease/model.py
```diff
@@ -56,7 +59,36 @@ def transition_names(self) -> List[TransitionString]:
             transitions += state.get_transition_names()
         return transitions
 
-    def setup(self, builder):
+    # @property
```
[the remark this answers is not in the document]
rmudambi: Almost. I need to flatten the commented out version. Thanks for catching.

mattkappel:
Looks like this needs linting

mattkappel (approved):

PR #350, opened 2023-09-12
refactor MSLT components

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: [MIC-4475](https://jira.ihme.washington.edu/browse/MIC-4475)

### Changes and notes
- Refactor MSLT components 

Note: this PR is huge, but don't worry about reviewing it too carefully. None of these components are used by any existing models. They may get used at some point in the future, but I suspect they didn't fully work before the refactor either.

### Testing
Automated tests pass

### DIFF
--- a/src/vivarium_public_health/mslt/delay.py
+++ b/src/vivarium_public_health/mslt/delay.py
@@ -7,11 +7,17 @@ This module contains tools to represent delayed effects in a multi-state
 lifetable simulation.
 
 """
+from typing import Any, Dict, List, Optional
+
 import numpy as np
 import pandas as pd
+from vivarium import Component
+from vivarium.framework.engine import Builder
+from vivarium.framework.event import Event
+from vivarium.framework.population import SimulantData
 
 
-class DelayedRisk:
+class DelayedRisk(Component):
     """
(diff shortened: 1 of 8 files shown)


src/vivarium_public_health/mslt/delay.py
```diff
@@ -7,11 +7,17 @@
...
 import numpy as np
 import pandas as pd
+from vivarium import Component
+from vivarium.framework.engine import Builder
+from vivarium.framework.event import Event
+from vivarium.framework.population import SimulantData
 
 
-class DelayedRisk:
+class DelayedRisk(Component):
```
albrja, line 20: What is the reasoning for making this a subclass of Component and not the VPH risk class?

albrja (approved):

src/vivarium_public_health/mslt/delay.py
rmudambi: It's an entirely different way of modelling risks that doesn't actually bear any resemblance to the standard VPH Risk component.

albrja (approved):

stevebachmeier (approved):

PR #352, opened 2023-09-13
Additional refactors to get CVD disease model working

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: [MIC-4534](https://jira.ihme.washington.edu/browse/MIC-4534)

### Changes and notes
- Reordered methods in `DiseaseState` to be better organized
- Moved the `add_dwell_time_transition()` method from `DiseaseState` to `BaseDiseaseState`
- Removed the use of `**kwargs` in the `__init__` methods of state and transition components as this was masking bugs
- Removed the use of `**kwargs` in the `add_transition` methods for the same reason.

### Testing
Ran US CVD for two timesteps
All automated tests pass

### DIFF
--- a/src/vivarium_public_health/mslt/disease.py
+++ b/src/vivarium_public_health/mslt/disease.py
@@ -7,11 +7,17 @@ This module contains tools for modeling diseases in multi-state lifetable
 simulations.
 
 """
+from typing import Any, Dict, List, Optional
+
 import numpy as np
 import pandas as pd
+from vivarium import Component
+from vivarium.framework.engine import Builder
+from vivarium.framework.event import Event
+from vivarium.framework.population import SimulantData
 
 
-class AcuteDisease:
+class AcuteDisease(Component):
     """
     An acute disease has a sufficiently short duration, relative to the
     time-step size, that it is not meaningful to talk about prevalence.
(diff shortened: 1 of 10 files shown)


src/vivarium_public_health/mslt/disease.py
```diff
@@ -96,33 +108,59 @@ class Disease:
...
+            self.disease: {
                 "simplified_no_remission_equations": False,
             },
         }
 
     @property
-    def name(self):
-        return self._name
+    def columns_created(self) -> List[str]:
+        columns = []
```
patricktnast, line 130: Nit: maybe something like
`## disease, rate, scenario, when`
`column_params = [[self.disease], ["_S", "_C"], ["", "_intervention"] , ["", "_previous"]]`
`return [''.join(x) for x in itertools.product(*column_params)]`
would be nice here?

patricktnast: I guess this is just a rearrangement of something that existed already, so maybe that's not on you to change here

patricktnast (approved):

src/vivarium_public_health/mslt/disease.py
rmudambi: That's probably better, but I was very intentionally trying to change as little as possible.

albrja (approved):

PR #366, opened 2023-10-05
Remove python3.8

hussain-jafari:
PyTables more or less deprecated Python 3.8 in their most recent release, and it's causing our builds to fail. We determined that it is probably best to just follow suit and drop 3.8 from our CI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11"]
     defaults:
       run:
         shell: bash -le {0}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
     defaults:
       run:
         shell: bash -le {0}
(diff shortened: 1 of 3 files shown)


patricktnast:
I'm not sure how much this matters, but we also have a hard-coded reference to python 3.8 in https://github.com/ihmeuw/vivarium_public_health/blob/65251452bb3e4826da165716140cd56947e97247/docs/source/conf.py#L198

patricktnast (approved):

CHANGELOG.rst
```diff
@@ -1,3 +1,7 @@
+**2.1.0 - 10/05/23**
+
+ - No longer using Python 3.8
```
rmudambi, line 3: Nit: I'd call this "Remove explicit support for Python 3.8"

.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11"]
```
rmudambi, line 18: Did you try adding 3.12 and see if it builds?

rmudambi (approved):

rmudambi:
You will also need to remove 3.8 (and add 3.12 if we go that route) to the status checks that need to pass in order to merge

rmudambi (approved):

CHANGELOG.rst
```diff
@@ -2,6 +2,10 @@
 
  - Remove explicit support for Python 3.8
 
+**2.0.2 - 10/05/23**
+
+ - Minor bugfix to ensure default remission rate calls the right artifact key
```
rmudambi, line 7: Did we actually release VPH 2.0.2? I didn't think we did. If we didn't, this information should be additional bullet point in the 2.1.0 release.

rmudambi (changes requested):

rmudambi (approved):

PR #367, opened 2023-10-09
remission rate bugfix changelog

hussain-jafari:
remission rate bugfix changelog


rmudambi:
This change should be reflected in https://github.com/ihmeuw/vivarium_public_health/pull/366, since changelog updates are per release, not per PR.

PR #368, opened 2023-10-13
method based disease setup

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4613

### Changes and notes
Perform BaseDiseaseState setup using more methods.

### Testing
Ran MNCH child sim for a few time steps.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ if __name__ == "__main__":
         long_description = f.read()
 
     install_requirements = [
-        "vivarium>=2.1.0",
+        "vivarium>=2.0.0",
         "numpy",
         "pandas",
         "scipy",
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ if __name__ == "__main__":
         long_description = f.read()
 
     install_requirements = [
-        "vivarium>=2.0.0",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -15,7 +15,7 @@
         long_description = f.read()
 
     install_requirements = [
-        "vivarium>=2.0.0",
+        "vivarium>=2.1.0",
```
rmudambi, line 18: Does this require vivarium 2.1.0 (i.e. is there some reason to believe it wouldn't be compatible with vivarium 2.0.1)? If not, we shouldn't touch this pin.

rmudambi (approved):

mattkappel (commented):
From what I can tell, this is mostly a refactor. Is that right? The changelog "Use methods to setup BaseDiseaseState" is a little vague.

mattkappel (approved):
Suggest adding more clarity to the changelog description

patricktnast (approved):

hussain-jafari:
@mattkappel How about something like "Perform actions in DiseaseState setup using class methods rather than hardcoding to allow for cleaner subclassing"?

PR #376, opened 2024-01-11
Introduce Causes Configuration Parser

rmudambi:
### Description

- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4608

### Changes and notes
- Create CausesConfigurationParser component that can parser a yaml 
  configuration of cause models into Vivarium components
- This was previously implemented in CVD and has been ported over and
  modified slightly (https://github.com/ihmeuw/vivarium_nih_us_cvd/blob/main/src/vivarium_nih_us_cvd/plugins/causes_parser.py)

N.B. for reviewers, if you want to see what is different from what was 
previously done in CVD you can exclude the first commit when reviewing.

### Testing
Developed automated tests of the new parser

### DIFF
--- a/src/vivarium_public_health/plugins/parser.py
+++ b/src/vivarium_public_health/plugins/parser.py
@@ -1,0 +1,358 @@
+from importlib import import_module
+from typing import Any, Callable, Dict, List, Union
+
+import pandas as pd
+from pkg_resources import resource_filename
+from vivarium import Component, ConfigTree
+from vivarium.framework.components import ComponentConfigurationParser
+from vivarium.framework.engine import Builder
+from vivarium.framework.state_machine import Trigger
+from vivarium_public_health.disease import (
+    BaseDiseaseState,
+    DiseaseModel,
+    DiseaseState,
+    RecoveredState,
+    SusceptibleState,
+    TransientDiseaseState,
+)
+from vivarium_public_health.utilities import TargetString
+
+from vivarium_nih_us_cvd.components.causes.state import (
+    MultiTransitionDiseaseState,
+    MultiTransitionSusceptibleState,
+)
+
+
+class CausesConfigurationParser(ComponentConfigurationParser):
+    """
+    Component configuration parser that acts the same as the standard vivarium
+    `ComponentConfigurationParser` but adds the additional ability to parse a
+    `causes` key and create `DiseaseModel` components.
+    """
+
+    def parse_component_config(self, component_config: ConfigTree) -> List[Component]:
+        """
+        Parses the component configuration and returns a list of components.
+
+        In particular, this method looks for an `external_configuration` key and
+        a `causes` key.
+
+        The `external_configuration` key should have names of packages that
+        contain cause model configuration files. Within that key should be a list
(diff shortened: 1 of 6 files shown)


src/vivarium_public_health/plugins/parser.py
```diff
@@ -85,93 +82,67 @@ def parse_component_config(self, component_config: ConfigTree) -> List[Component
...
-
-        components += self.process_level(component_config_dict, [])
-        return components
-
-    ##################################
-    # Configuration creation methods #
-    ##################################
-
-    @staticmethod
-    def mark_multi_transition_states(component_config: ConfigTree) -> None:
```
stevebachmeier: Are we dropping this feature altogether or is it implemented elsewhere now?

src/vivarium_public_health/plugins/parser.py
```diff
@@ -209,7 +180,14 @@ def get_cause_model_components(self, causes_config: ConfigTree) -> List[Componen
                     transition_config,
                 )
 
-            cause_models.append(DiseaseModel(cause_name, states=list(states.values())))
+            model_type = import_by_path(cause_config.model_type)
+            initial_state = (
+                states[cause_config.initial_state] if cause_config.initial_state else None
```
stevebachmeier, line 185: nit: `states.get[cause_config.initial_state], None`

tests/conftest.py
```diff
@@ -23,24 +24,32 @@ def pytest_collection_modifyitems(config, items):
             item.add_marker(skip_slow)
 
 
-@pytest.fixture(scope="function")
-def base_config():
-    config = build_simulation_configuration()
+@pytest.fixture(scope="session")
+def base_config_factory() -> Callable[[], ConfigTree]:
+    def _base_config() -> ConfigTree:
```
stevebachmeier, line 29: Why the misdirection?

stevebachmeier (approved):

src/vivarium_public_health/plugins/parser.py
rmudambi: I'm dropping this entirely, as I don't think it has any reason to exist now that we have individual clocks.

tests/conftest.py
rmudambi: I think you mean "indirection", and this is because I have module scoped fixtures that use this fixture, and you can't pass a fixture with a smaller scope as an argument to a fixture with a larger scope. From my reading, it seems fixtures is the correct way to handle this.

src/vivarium_public_health/plugins/parser.py
rmudambi: I thought about that, but that doesn't actually have the same behavior, since it will return None if the `initial_state` is not `None`, but not present in the list of states for the model, whereas right now that will throw a `KeyError`. If we have proper validation of the config (which it should, and will happen before this), we should error before getting to this point in that case. I will put a comment here and make your suggested change in the validation PR.

albrja (approved):

hussain-jafari (approved):

patricktnast (approved):

PR #381, opened 2024-02-14
Bugfix/sbachmei/mic 4886 fix one year assign demographic proportions

stevebachmeier:
### Description

- *Category*:  bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4886

### Changes and notes
 Syl found when archiving the CVD project with artifacts subset to only 
2019 data that one of the data transformations is broken. Specifically, the lambda
function being applied to a single group returns a wide one-row dataframe
instead of the expected pd.Series. 

(https://stackoverflow.com/questions/69109980/unclear-why-groupby-with-single-group-produces-row-dataframe)

### Testing

Ran the cvd simulation with both the subset one-year artifacts and the old
multi-year artifacts; both simulations actually started and finished the
first time step.

### DIFF
--- a/src/vivarium_public_health/population/data_transformations.py
+++ b/src/vivarium_public_health/population/data_transformations.py
@@ -7,6 +7,7 @@ This module contains tools for handling raw demographic data and transforming
 it into different distributions for sampling.
 
 """
+
 from collections import namedtuple
 from typing import Tuple, Union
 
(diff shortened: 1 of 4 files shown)


tests/population/test_data_transformations.py
```diff
@@ -104,9 +124,9 @@ def test_smooth_ages(pop_data, include_sex):
     simulants = pd.DataFrame(
         {
             "age": [22.5] * 10000 + [52.5] * 10000,
-            "sex": ["Male", "Female"] * 10000
-            if include_sex == "Both"
-            else [include_sex] * 20000,
+            "sex": (
```
stevebachmeier, line 127: This is just black doing its thing

.github/CODEOWNERS
```diff
@@ -1,2 +1,2 @@
 # default owners
-* @albrja @collijk @hussain-jafari @mattkappel @patricktnast @ramittal @rmudambi @stevebachmeier
+* @albrja @collijk @hussain-jafari @patricktnast @rmudambi @stevebachmeier
```
stevebachmeier, line 2: I don't know why this is part of my merge but it looks correct ¯\_(ツ)_/¯

rmudambi: I think it's because you branched off of main and I don't have that change in my epic branch.

rmudambi (approved):

.github/CODEOWNERS
stevebachmeier: Ah, yeah, def that. I'm so used to rebasing I forgot that this sort of things happen when you merge.

tests/population/test_data_transformations.py
```diff
@@ -48,7 +62,13 @@ def test_assign_demographic_proportions(pop_data, include_sex):
         )
 
 
-def test_rescale_binned_proportions_full_range(pop_data, include_sex):
+def test_single_group_assign_demographic_proportions(pop_data_single_group):
```
stevebachmeier, line 65: I tried implementing this as a parameterization on the previous test but it got complicated w/ `include_sex` which is in turn a parameter. So I just made this separate/simple test

src/vivarium_public_health/population/data_transformations.py
```diff
@@ -54,23 +55,23 @@ def assign_demographic_proportions(
 
     population_data["P(sex, location, age| year)"] = (
         population_data.groupby("year_start", as_index=False)
-        .apply(lambda sub_pop: sub_pop.value / sub_pop.value.sum())
-        .reset_index(level=0)
-        .value.fillna(0.0)
+        .apply(lambda sub_pop: sub_pop[["value"]] / sub_pop["value"].sum())
+        .reset_index(level=0)["value"]
```
stevebachmeier, line 59: I updated the value column dot-notation (`.value`) with bracket notation (`["value"]`) for consistency w/ how we usually do things.

patricktnast (approved):

PR #385, opened 2024-03-05
Mic-4591/all-deaths-switch

albrja:
### Adds ability for Mortality Observer to aggregate deaths and yllls or record at cause specific level.
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4591

### Changes and notes
-adds configuration key "aggregate" to mortality observer
-allows mortality observer to record total deaths and ylls instead of cause specific to improve runtime

### Testing
Ran simulation with both the aggregate True and False and saw expected behavior.

### DIFF
--- a/src/vivarium_public_health/metrics/mortality.py
+++ b/src/vivarium_public_health/metrics/mortality.py
@@ -7,7 +7,7 @@ This module contains tools for observing cause-specific and
 excess mortality in the simulation, including "other causes".
 
 """
-from typing import List, Optional
+from typing import Callable, List, Optional
 
 import pandas as pd
 from vivarium import Component
@@ -22,7 +22,9 @@ class MortalityObserver(Component):
     By default, this counts cause-specific deaths and years of life lost over
     the full course of the simulation. It can be configured to add or remove
     stratification groups to the default groups defined by a
-    :class:ResultsStratifier.
+    :class:ResultsStratifier. The aggregate configuration key can be set to
+    True to aggregate all deaths and ylls into a single observation and remove
+    the stratification by cause of death to improve runtime.
 
     In the model specification, your configuration for this component should
     be specified as, e.g.:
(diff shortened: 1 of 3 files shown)


albrja:
I feel like it looks like there is more code being changed than there was but I also just wrapped `builder.results.register_observation` so that is probably why. Anyways the take away from this is you now have a new "aggregate" configuration key which defaults to False. This will key the current behavior of the mortality observer.

However, if you change that key to True in the yaml, you then will only register observations for total deaths and ylls (and in turn not have cause specific metrics). This will allow us to save significant runtime when we do not care about the cause specific metrics and only total counts.

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -78,40 +81,84 @@ def setup(self, builder: Builder) -> None:
         self.causes_of_death = ["other_causes"] + [
```
rmudambi, line 81: I believe this is only used if aggregation is false, so you could move it inside the if block

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -78,40 +81,84 @@ def setup(self, builder: Builder) -> None:
...
                     "exit_time",
                     "years_of_life_lost",
                 ],
-                additional_stratifications=self.config.include,
-                excluded_stratifications=self.config.exclude,
-                when="collect_metrics",
             )
 
+    def _register_aggregate_observations(self, builder: Builder) -> None:
+        self._register_observation(
```
rmudambi, line 116: I was envisioning a different way to reduce code duplication. Let's talk about it after standup

tests/metrics/test_mortality_observer.py
```diff
@@ -156,3 +156,51 @@ def test_observation_correctness(simulation_after_one_step):
         assert np.isclose(
             results(pop.index)[observation], expected_ylls[observation], rtol=0.001
         )
+
+
+def test_aggregation_configruation(base_config, base_plugins):
```
rmudambi, line 161: Nit: typo - configruation

patricktnast (approved):

hussain-jafari (approved):

PR #394, opened 2024-05-01
Improve Risk component LookupTable configuration

rmudambi:
### Description

- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5024

### Changes and notes
- Update risk lookup table configurations to match new usage
- Adds some typing 

### Testing
- All tests pass
- Verified manually that categorical and ensemble risks work as expected.

### DIFF
--- a/src/vivarium_public_health/risks/base_risk.py
+++ b/src/vivarium_public_health/risks/base_risk.py
@@ -157,6 +157,7 @@ class Risk(Component):
     #################
 
     def build_all_lookup_tables(self, builder: "Builder") -> None:
+        # exposure lookup tables are handled by the SimulationDistribution subcomponent
         pass
 
     def get_randomness_stream(self, builder: Builder) -> RandomnessStream:
--- a/src/vivarium_public_health/risks/base_risk.py
+++ b/src/vivarium_public_health/risks/base_risk.py
@@ -87,21 +87,18 @@ class Risk(Component):
     ##############
 
     @property
+    def name(self) -> str:
+        return self.risk
+
+    @property
     def configuration_defaults(self) -> Dict[str, Any]:
         return {
-            self.risk.name: {
-                "exposure": self.build_lookup_table_config(
-                    value="data",
-                    continuous_columns=["age", "year"],
-                    categorical_columns=["sex"],
-                    key_name=f"risk_factor.{self.risk.name}.exposure",
-                ),
-                "ensemble_distribution_weights": {
-                    "key_name": f"risk_factor.{self.risk.name}.exposure_distribution_weights",
+            self.name: {
+                "data_sources": {
+                    "exposure": f"{self.risk}.exposure",
                 },
-                "exposure_standard_deviation": {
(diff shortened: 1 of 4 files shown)


src/vivarium_public_health/risks/base_risk.py
```diff
@@ -86,22 +86,19 @@ class Risk(Component):
...
-                "ensemble_distribution_weights": {
-                    "key_name": f"risk_factor.{self.risk.name}.exposure_distribution_weights",
-                },
-                "exposure_standard_deviation": {
-                    "key_name": f"risk_factor.{self.risk.name}.exposure_standard_deviation",
+            self.name: {
+                "data_sources": {
+                    "exposure": f"{self.risk}.exposure",
                 },
+                "ensemble_distribution_weights": f"{self.risk}.exposure_distribution_weights",
```
albrja, line 100: Shouldn't these two be in the data sources?

src/vivarium_public_health/risks/base_risk.py
```diff
@@ -158,6 +156,9 @@ def get_exposure_distribution(self) -> SimulationDistribution:
     # Setup methods #
     #################
 
+    def build_lookup_tables(self, builder: "Builder") -> None:
```
albrja, line 159: For the other reviewers I found this confusing at first - this is overwriting the component method it inherits to do nothing and then the subcomponent Risk creates is overwriting this method again.

src/vivarium_public_health/risks/data_transformations.py
```diff
@@ -409,23 +438,23 @@ def validate_distribution_data_source(builder, risk: EntityString):
...
         raise ValueError(f"Unknown risk type {risk.type} for risk {risk.name}")
 
 
+def validate_distribution_data(distribution_data: Dict[str, Any]) -> None:
+    exposure_data = distribution_data["exposure"]
+    val_cols = distribution_data["exposure_value_columns"]
+    if distribution_data["distribution_type"] == "dichotomous":
+        if ((exposure_data[val_cols] < 0) | exposure_data[val_cols] > 1).any().any():
+            raise ValueError(f"Exposure should be in the range [0, 1]")
+    # TODO: validate that weights, standard deviation, and exposure have the
```
albrja, line 451: Not in this epic I assume?

src/vivarium_public_health/risks/base_risk.py
rmudambi: Oh, good catch. It doesn't make a difference as it turns out, since we don't iterate through these keys, but they should be there for consistency.

src/vivarium_public_health/risks/data_transformations.py
rmudambi: We probably should do this while we're here, but after we're done with everything that is required for the MNCH model

stevebachmeier (approved):

albrja (approved):

hussain-jafari (approved):

PR #396, opened 2024-05-02
create cause model lookup tables via configuration

rmudambi:
### Description

- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5003

### Changes and notes
- Enable creation of all cause model lookup tables using Component inheritance 

### Testing
All automated tests pass

### DIFF
--- a/src/vivarium_public_health/disease/model.py
+++ b/src/vivarium_public_health/disease/model.py
@@ -9,7 +9,7 @@ transitions at simulation initialization and during transitions.
 
 """
 
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
@@ -137,8 +137,6 @@ class DiseaseModel(Machine):
 
         assert self.initial_state in {s.state_id for s in self.states}
 
-        # FIXME: this is a hack to figure out whether or not we're at the simulation start based on the fact that the
-        #  fertility components create this user data
         if pop_data.user_data["sim_state"] == "setup":  # simulation start
             if self.configuration_age_start != self.configuration_age_end != 0:
                 state_names, weights_bins = self.get_state_weights(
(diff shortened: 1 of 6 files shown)


src/vivarium_public_health/disease/model.py
```diff
@@ -132,8 +137,6 @@ def on_initialize_simulants(self, pop_data: SimulantData) -> None:
 
         assert self.initial_state in {s.state_id for s in self.states}
 
-        # FIXME: this is a hack to figure out whether or not we're at the simulation start based on the fact that the
-        #  fertility components create this user data
```
stevebachmeier, line 136: Is this just no longer a hack?

src/vivarium_public_health/disease/transition.py
```diff
@@ -28,6 +30,35 @@ def __new__(cls, value):
...
+        if "incidence_rate" in self._get_data_functions:
+            pipeline_name = f"{self.output_state.state_id}.incidence_rate"
+        elif "remission_rate" in self._get_data_functions:
+            pipeline_name = f"{self.input_state.state_id}.remission_rate"
+        elif "transition_rate" in self._get_data_functions:
+            pipeline_name = (
+                f"{self.input_state.state_id}_to_{self.output_state.state_id}.transition_rate"
+            )
+        else:
+            raise ValueError("Cannot determine rate_transition pipeline name.")
```
stevebachmeier, line 59: Is there a way to make this more informative for the user? Maybe even just saying something like "expected to find 'incidence_rate', 'remission_rate', or 'transition_rate' but was provided {list(get_data_functions.keys())}"

src/vivarium_public_health/disease/transition.py
```diff
@@ -48,19 +79,16 @@ def __init__(
...
-            rate_data, key_columns=["sex"], parameter_columns=["age", "year"]
-        )
+        lookup_columns = get_lookup_columns([self.lookup_tables["transition_rate"]])
         self.transition_rate = builder.value.register_rate_producer(
-            pipeline_name,
+            self.transition_rate_pipeline_name,
             source=self.compute_transition_rate,
-            requires_columns=["age", "sex", "alive"],
-            requires_values=[f"{pipeline_name}.paf"],
+            requires_columns=lookup_columns + ["alive"],
```
stevebachmeier, line 86: Why is 'alive' not just part of the `get_lookup_columns` return?

stevebachmeier (approved):

src/vivarium_public_health/disease/model.py
rmudambi: I don't know if it's a hack or not, but it's been around for 5+ years. It doesn't seem terrible and clearly it's not a priority to do anything about it, in any case.

src/vivarium_public_health/disease/transition.py
rmudambi: 'alive' is not a column that the LookupTable uses. It is needed by the pipeline because the pipeline source uses it separately to filter the data.

albrja (approved):

PR #400, opened 2024-05-07
Feature/sbachmei/mic 4981 5036 update disability observer

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5036

### Changes and notes
 
Updates the DisabilityObserver to inherit from the new StratifiedObserver
class as well as the new results processing system.

### Testing

tests pass

### DIFF
--- a/src/vivarium_public_health/metrics/disability.py
+++ b/src/vivarium_public_health/metrics/disability.py
@@ -7,11 +7,13 @@ This module contains tools for observing years lived with disability (YLDs)
 in the simulation.
 
 """
+
 from typing import List
 
 import pandas as pd
 from vivarium import Component
 from vivarium.framework.engine import Builder
+from vivarium.framework.results.reporters import aggregate_dataframes_to_csv
 from vivarium.framework.values import Pipeline, list_combiner, union_post_processor
 
 from vivarium_public_health.disease import DiseaseState, RiskAttributableDisease
@@ -80,6 +82,7 @@ class DisabilityObserver(Component):
             additional_stratifications=self.config.include,
             excluded_stratifications=self.config.exclude,
             when="time_step__prepare",
+            report=aggregate_dataframes_to_csv,
         )
 
         for cause_state in cause_states:
@@ -96,6 +99,7 @@ class DisabilityObserver(Component):
                 additional_stratifications=self.config.include,
                 excluded_stratifications=self.config.exclude,
                 when="time_step__prepare",
+                report=aggregate_dataframes_to_csv,
             )
 
     #################
--- a/src/vivarium_public_health/metrics/disability.py
+++ b/src/vivarium_public_health/metrics/disability.py
@@ -12,10 +12,10 @@ from pathlib import Path
 from typing import List
 
 import pandas as pd
-
 from vivarium.framework.engine import Builder
 from vivarium.framework.results import METRICS_COLUMN, StratifiedObserver
 from vivarium.framework.values import Pipeline, list_combiner, union_post_processor
+
 from vivarium_public_health.disease import DiseaseState, RiskAttributableDisease
 from vivarium_public_health.utilities import to_years
 
--- a/src/vivarium_public_health/metrics/disability.py
+++ b/src/vivarium_public_health/metrics/disability.py
@@ -14,7 +14,7 @@ from typing import List
 import pandas as pd
 
 from vivarium.framework.engine import Builder
-from vivarium.framework.results.observer import StratifiedObserver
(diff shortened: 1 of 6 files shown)


tests/metrics/test_disability_observer.py
```diff
@@ -194,21 +196,43 @@ def test_disability_accumulation(
...
+        itertools.product(*[["Female", "Male"], list(yld_masks)])
+    )
+
+    # Check other columns
+    assert set(results.columns) == set(
+        ["sex", "cause", "measure", "random_seed", "input_draw", "value"]
+    )
+    assert (results["measure"] == "ylds").all()
+    assert (results["random_seed"] == 0).all()
+    assert (results["input_draw"]).isna().all()
```
stevebachmeier, line 224: From @hussain-jafari in the old PR (that had the ref HEAD borked and so wasn't passing builds):

We had a short discussion about this at mid-day but want to confirm that we want this be to NA

stevebachmeier: > From @hussain-jafari in the old PR (that had the ref HEAD borked and so wasn't passing builds):
> 
> > We had a short discussion about this at mid-day but want to confirm that we want this be to NA

I think it's fine. The fact is, the test is such that input_draw _is_ missing.

src/vivarium_public_health/metrics/disability.py
```diff
@@ -116,3 +113,36 @@ def get_disability_weight_pipeline(self, builder: Builder) -> Pipeline:
...
+    def report(
+        self,
+        measure: str,
+        results: pd.DataFrame,
+    ) -> None:
+        """Combine the measure-specific observer results and save to a single file."""
+        results_dir = Path(self.results_dir)
+        measure, cause = measure.split("due_to")
+        measure = measure.strip("_")
+        cause = cause.strip("_")
```
rmudambi, line 130: Nit: Could do
```suggestion
        measure, cause = [s.strip("_") for s in measure.split("due_to")]
```

src/vivarium_public_health/metrics/disability.py
```diff
@@ -116,3 +113,36 @@ def get_disability_weight_pipeline(self, builder: Builder) -> Pipeline:
...
+        # Add extra cols
+        results["measure"] = measure
+        results["cause"] = cause
+        results["random_seed"] = self.random_seed
```
rmudambi, line 134: Maybe only add these columns if they are not None

src/vivarium_public_health/metrics/disability.py
```diff
@@ -116,3 +113,36 @@ def get_disability_weight_pipeline(self, builder: Builder) -> Pipeline:
...
+        results["input_draw"] = self.input_draw
+        # Sort the columns such that the stratifications (index) are first
+        # and "value" is last and sort the rows by the stratifications.
+        other_cols = [c for c in results.columns if c != "value"]
```
rmudambi, line 138: How do we know the columns we're looking for is called "value"? Where is this defined? Can there be a constant somewhere that is referenced?

src/vivarium_public_health/metrics/disability.py
```diff
@@ -116,3 +113,36 @@ def get_disability_weight_pipeline(self, builder: Builder) -> Pipeline:
...
+        results = results[other_cols + ["value"]].sort_index().reset_index()
+
+        # Concat and save
+        results_file = results_dir / f"{measure}.csv"
```
rmudambi, line 142: Is it possible that different observers could have the same measure (and therefore try to write to the same file?

rmudambi (approved):

src/vivarium_public_health/metrics/disability.py
stevebachmeier: [It's defined in vivarium](https://github.com/ihmeuw/vivarium/blob/epic/MIC-4968-results-processing/src/vivarium/framework/results/context.py#L138). I'll make a constant

src/vivarium_public_health/metrics/disability.py
stevebachmeier: Hmm, so the measure is actually just the observation name so I guess if you registered more than one observer w/ the same name that would be a problem. I'll create a ticket to check for duplicate names and raise if so

stevebachmeier: But that nuance aside, yes this could be a problem. Is it a realistic thing though to have different observers observing the same thing? I'm not sure how offhand to check for this.

PR #401, opened 2024-05-08
Feature/sbachmei/mic 4981 5037 update disease observer

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5037

### Testing

Tests pass

### DIFF
--- a/src/vivarium_public_health/metrics/disease.py
+++ b/src/vivarium_public_health/metrics/disease.py
@@ -7,18 +7,20 @@ This module contains tools for observing disease incidence and prevalence
 in the simulation.
 
 """
+
+from pathlib import Path
 from typing import Any, Dict, List, Optional
 
 import pandas as pd
-from vivarium import Component
+
 from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
 from vivarium.framework.population import SimulantData
-
+from vivarium.framework.results import METRICS_COLUMN, StratifiedObserver
 from vivarium_public_health.utilities import to_years
 
 
-class DiseaseObserver(Component):
+class DiseaseObserver(StratifiedObserver):
     """Observes disease counts and person time for a cause.
 
     By default, this observer computes aggregate disease state person time and
@@ -40,15 +42,6 @@ class DiseaseObserver(Component):
                         - "sample_stratification"
     """
 
-    CONFIGURATION_DEFAULTS = {
-        "stratification": {
-            "disease": {
-                "exclude": [],
-                "include": [],
-            }
-        }
-    }
-
     ##############
     # Properties #
     ##############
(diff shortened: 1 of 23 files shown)


src/vivarium_public_health/metrics/disease.py
```diff
@@ -57,7 +50,7 @@ class DiseaseObserver(Component):
     def configuration_defaults(self) -> Dict[str, Any]:
         return {
             "stratification": {
-                self.disease: self.CONFIGURATION_DEFAULTS["stratification"]["disease"]
+                self.disease: super().configuration_defaults["stratification"]["disease"]
```
rmudambi, line 53: Why do we need this? Isn't just using the superclass `StratifiedObserver` sufficient?

src/vivarium_public_health/metrics/disease.py
```diff
@@ -134,3 +133,53 @@ def on_time_step_prepare(self, event: Event) -> None:
 
     def aggregate_state_person_time(self, x: pd.DataFrame) -> float:
         return len(x) * to_years(self.step_size())
+
+    ##################
+    # Report methods #
+    ##################
+
+    def report_person_time(self, measure: str, results: pd.DataFrame):
```
rmudambi, line 141: It seems like there's a lot of duplicated code between these two report methods and also the report method of the `DisabilityObserver`

stevebachmeier: I'll look and see if I can extract something out into a utility function

src/vivarium_public_health/metrics/disease.py
stevebachmeier: No. The super() would return
```
"stratification": {
  "disease": {
    "exclude": [],
    "include": [],
  },
}
```

But we don't want a generic "disease" for the second key, but the specific disease name

src/vivarium_public_health/utilities.py
```diff
@@ -123,3 +126,32 @@ def get_index_columns_from_lookup_configuration(
...
+    results_dir = Path(observer.results_dir)
+    # Add extra cols
+    col_mapper = {"measure": measure}
+    col_mapper.update(extra_cols)
+    col_mapper.update(
+        {"random_seed": observer.random_seed, "input_draw": observer.input_draw}
+    )
+    for col, val in col_mapper.items():
+        if val is not None:
+            results[col] = val
```
rmudambi, line 147:
```suggestion
    col_mapper = {k: v for k, v in col_mapper.items() if v is not None}
```

stevebachmeier: Decided not to change. It's the same number of lines and I think it's more readable (despite hating nested loops)

rmudambi (approved):

PR #409, opened 2024-05-10
Feature/sbachmei/reformat output datasets

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: na

### Changes and notes
 
Instead of saving all measure results in a single file regardless of
where they came from (e.g. transition counts and person time from
different diseases or person time from diseases and risks), we
want to save each specific observer as its own dataset (e.g.
each disease gets its own specific person_time_{disease}.parquet
and transtion_count_{disease}.parquet.

This also streamlines the datasets so that they always have `measure`,
`entity_type`, `entity`, and `sub_entity` columns per RT's [design doc](https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_features/standard_results_formats/index.html)

### Testing

all tests pass

### DIFF
--- a/src/vivarium_public_health/metrics/disability.py
+++ b/src/vivarium_public_health/metrics/disability.py
@@ -126,22 +126,17 @@ class DisabilityObserver(StratifiedObserver):
         results: pd.DataFrame,
     ) -> None:
         """Combine each observation's results and save to a single file"""
-        measure = "ylds"
-        if not cause_state:
-            entity_type = "cause"
-            entity = "all_causes"
-            sub_entity = None
-        else:
-            entity_type = cause_state.cause_type
-            entity = cause_state.model
-            sub_entity = cause_state.state_id
+
+        kwargs = {
+            "entity_type": cause_state.cause_type if cause_state else "cause",
+            "entity": cause_state.model if cause_state else "all_causes",
+            "sub_entity": cause_state.state_id if cause_state else None,
+            "results_dir": self.results_dir,
+            "random_seed": self.random_seed,
+            "input_draw": self.input_draw,
+        }
         write_dataframe_to_parquet(
             results=results,
-            measure=measure,
-            entity_type=entity_type,
-            entity=entity,
-            sub_entity=sub_entity,
-            results_dir=self.results_dir,
-            random_seed=self.random_seed,
-            input_draw=self.input_draw,
+            measure="ylds",
+            **kwargs,
         )
--- a/src/vivarium_public_health/metrics/disability.py
+++ b/src/vivarium_public_health/metrics/disability.py
@@ -8,7 +8,8 @@ in the simulation.
 
 """
 
-from typing import List
+from functools import partial
+from typing import List, Optional
 
 import pandas as pd
 from vivarium.framework.engine import Builder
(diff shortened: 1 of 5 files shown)


src/vivarium_public_health/metrics/disability.py
```diff
@@ -120,16 +121,27 @@ def disability_weight_aggregator(self, dw: pd.DataFrame) -> float:
 
     def report(
         self,
+        cause_state: Optional[DiseaseState],
```
albrja, line 124: Doesn't the optional need to be the last arg or no?

albrja (approved):

src/vivarium_public_health/metrics/disability.py
stevebachmeier: Nope, Optional inplicitly allows for None type. That is, `Optional[DiseaseState]` == `Union[DiseaseState, None]`

src/vivarium_public_health/metrics/disability.py
```diff
@@ -120,16 +121,27 @@ def disability_weight_aggregator(self, dw: pd.DataFrame) -> float:
...
-            self.input_draw,
-            {COLUMNS.CAUSE: cause},
+            results=results,
+            measure=measure,
+            entity_type=entity_type,
+            entity=entity,
+            sub_entity=sub_entity,
+            results_dir=self.results_dir,
+            random_seed=self.random_seed,
+            input_draw=self.input_draw,
```
rmudambi, line 146: Is it cleaner to build a dictionary `kwargs` and then do `write_dataframe_to_parquet(**kwargs)`?

src/vivarium_public_health/metrics/disability.py
```diff
@@ -120,16 +121,27 @@ def disability_weight_aggregator(self, dw: pd.DataFrame) -> float:
 
     def report(
```
rmudambi, line 122: I think it's a bit confusing to call this method `report`, since it's not actually the method being sent as the report function when you register observations. Same thing in your other observer.

tests/metrics/test_disease_observer.py
```diff
@@ -218,3 +227,58 @@ def test_observation_correctness(base_config, base_plugins, disease, model, tmpd
...
+    hcd_healthy_state = SusceptibleState("not_a_zombie")
+    hcd_infected_state = DiseaseState("a_zombie")
+    hcd_healthy_state.add_rate_transition(hcd_infected_state)
+    human_cortico_deficiency = DiseaseModel(
+        "human_cortico_deficiency",
+        initial_state=hcd_healthy_state,
+        states=[hcd_healthy_state, hcd_infected_state],
+    )
+
+    vampiris_observer = DiseaseObserver("vampiris")
```
rmudambi, line 251: Vampiris not vampirism?

rmudambi (approved):

tests/metrics/test_disease_observer.py
stevebachmeier: https://imlegend.fandom.com/wiki/Vampiris

src/vivarium_public_health/metrics/disability.py
stevebachmeier: Not sure I understand. In line 100, we register the observation with `report=partial(self.report, cause_state)`. The self.report is indeed this `report` method defined here.

src/vivarium_public_health/metrics/disability.py
stevebachmeier: I'm pretty indifferent. Slightly less lines, though I find inline if/else statements harder to read than actual if: else: blocks

PR #414, opened 2024-05-15
update LBWSG parsing

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4997

### Changes and notes
Account for + in right interval of some ages and closed bracket in right interval of some weights by adding more splitting. 

### Testing
Ran through all category descriptions in LBWSG risk factor in gbd_mapping and checked that parse_description properly parsed these descriptions.

### DIFF
--- a/src/vivarium_public_health/risks/implementations/low_birth_weight_and_short_gestation.py
+++ b/src/vivarium_public_health/risks/implementations/low_birth_weight_and_short_gestation.py
@@ -192,10 +192,12 @@ class LBWSGDistribution(PolytomousDistribution):
         """
         endpoints = {
             BIRTH_WEIGHT: [
-                float(val) for val in description.split(", [")[1].split(")")[0].split(']')[0].split(", ")
+                float(val)
+                for val in description.split(", [")[1].split(")")[0].split("]")[0].split(", ")
             ],
             GESTATIONAL_AGE: [
-                float(val) for val in description.split("- [")[1].split(")")[0].split('+')[0].split(", ")
+                float(val)
+                for val in description.split("- [")[1].split(")")[0].split("+")[0].split(", ")
             ],
         }[axis]
         return pd.Interval(*endpoints, closed="left")  # noqa
--- a/src/vivarium_public_health/risks/implementations/low_birth_weight_and_short_gestation.py
+++ b/src/vivarium_public_health/risks/implementations/low_birth_weight_and_short_gestation.py
@@ -153,6 +153,13 @@ class LBWSGDistribution(PolytomousDistribution):
         """
         Parses a string corresponding to a low birth weight and short gestation
         category to an Interval.
+        An example of a standard description:
+        'Neonatal preterm and LBWSG (estimation years) - [0, 24) wks, [0, 500) g'
+        An example of an edge case for gestational age:
+        'Neonatal preterm and LBWSG (estimation years) - [40, 42+] wks, [2000, 2500) g'
+        An example of an edge case of birth weight:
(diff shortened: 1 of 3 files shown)


stevebachmeier (approved):
I'm trusting you testing this well. What does `descripion` look like?

hussain-jafari:
> I'm trusting you testing this well. What does `description` look like?

We read in description from the risk_factor.low_birth_weight_and_short_gestation.categories key in the artifact, which uses load_metadata to read the LBWSG categories from gbd_mapping and create a dictionary with entries that look like {'cat2': 'Birth prevalence - [0, 24) wks, [0, 500) g'}.

src/vivarium_public_health/risks/implementations/low_birth_weight_and_short_gestation.py
```diff
@@ -152,17 +152,26 @@ def single_axis_ppf(
...
+
         :param axis:
         :param description:
         :return:
         """
         endpoints = {
             BIRTH_WEIGHT: [
-                float(val) for val in description.split(", [")[1].split(")")[0].split(", ")
+                float(val)
+                for val in description.split(", [")[1].split(")")[0].split("]")[0].split(", ")
```
stevebachmeier, line 170: I'd consider using stricter regex for this to extract the values between [ or ( and ] or ). This will only work if you've provided in your examples all edge cases, though. Specifically,

- The age is always before the weight (this is an inherent assumption with the current approach as well)
- The age and weight are always "wks" and "g" and separated by ", " (including the space).
- They are always surrounded by [] or () (or any combination of)
- The two values in each set are always separated by  ", " (including the space)
- Each value only consists of digits except for the second age value which can optionally include a "+"
- Anything prior to the age and weight never look like one of these intervals (i.e. while "estimation years" is in parentheses, it's NOT two digits separated by ", " so it doesn't match

idk, maybe this isn't worth the squeeze. REgardless, I agree you need tests 

```
>>> import re
>>>
>>> pattern = r'[\[\(](\d+, \d+\+?)[\]\)] wks, [\[\(](\d+, \d+)[\]\)] g'
>>>
>>> text1 = 'Neonatal preterm and LBWSG (estimation years) - [0, 24) wks, [0, 500) g'
>>> match = re.search(pattern, text1)
>>> age_range = match.group(1)
>>> age_range
'0, 24'
>>> weight_range = match.group(2)
>>> weight_range
'0, 500'
>>> 
>>>
>>>
>>> text2 = 'Neonatal preterm and LBWSG (estimation years) - [40, 42+] wks, [2000, 2500) g'
>>> match = re.search(pattern, text2)
>>> age_range = match.group(1)
>>> age_range
'40, 42+'
>>> weight_range = match.group(2)
>>> weight_range
'2000, 2500'
>>> 
>>> 
>>> 
>>> text3 = 'Neonatal preterm and LBWSG (estimation years) - [36, 37) wks, [4000, 9999] g'
>>> match = re.search(pattern, text3)
>>> age_range = match.group(1)
>>> age_range
'36, 37'
>>> weight_range = match.group(2)
>>> weight_range
'4000, 9999'
```

stevebachmeier: Another potentally cleaner solution is similar to what you've done but relying more on `strip`. This splits on "wks", replaces all brackets or parentheses w/ emtpy strings, and then strips excess white space

```
age_range = re.sub(r"[\[\]\(\)]", "", text1.split("wks")[0].split(" - ")[1]).strip()
```
and something similar for weight

rmudambi (approved):

PR #419, opened 2024-05-23
register one observation each for deaths and ylls

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5062

### Changes and notes
This is one of two PRs to refactor Mortality (this one) and Disability 
Observers. Those are currently registering one observation per
cause where we really want to register a single observation
stratified by cause.

This is the easier of the two: it essentially adds cause_of_death
as a stratification (outside of ResultsStratifier, though, as only
the Mortality component needs it) and then handles the
"all_causes" special case manually.

### Testing

tests pass

### DIFF
--- a/src/vivarium_public_health/metrics/mortality.py
+++ b/src/vivarium_public_health/metrics/mortality.py
@@ -8,15 +8,14 @@ excess mortality in the simulation, including "other causes".
 
 """
 
-from functools import partial
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional
 
 import pandas as pd
 from vivarium.framework.engine import Builder
 from vivarium.framework.results import StratifiedObserver
 
 from vivarium_public_health.disease import DiseaseState, RiskAttributableDisease
-from vivarium_public_health.metrics.reporters import write_dataframe_to_parquet
+from vivarium_public_health.metrics.reporters import COLUMNS, write_dataframe
 
 
 class MortalityObserver(StratifiedObserver):
@@ -49,7 +48,7 @@ class MortalityObserver(StratifiedObserver):
 
     def __init__(self) -> None:
         super().__init__()
-        self.required_death_columns = ["alive", "exit_time"]
+        self.required_death_columns = ["alive", "exit_time", "cause_of_death"]
         self.required_yll_columns = [
             "alive",
             "cause_of_death",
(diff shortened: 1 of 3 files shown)


src/vivarium_public_health/metrics/reporters.py
```diff
@@ -26,6 +26,18 @@ def name(self) -> str:
...
+    measure: str,
+    results_dir: Optional[Union[str, Path]],
+) -> None:
+    """Utility function for observation 'report' methods to write pd.DataFrames to parquet"""
+    if results_dir is None:
+        raise ValueError("A results_dir must be specified to write out results.")
+    results.to_parquet(Path(results_dir) / f"{measure}.parquet", index=False)
+
+
+# FIXME: remove this when ready
```
stevebachmeier, line 40: I need to update the other components to use the new `write_dataframe` method and then I'll remove this

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -92,61 +91,44 @@ def columns_required(self) -> Optional[List[str]]:
     def setup(self, builder: Builder) -> None:
         self.clock = builder.time.clock()
         self.config = builder.configuration.stratification.mortality
+        self.causes_of_death = [
```
albrja, line 94: I think this could have stayed in the other method - Im not sure it matters but it made it confusing to read.

albrja (approved):

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -167,23 +149,50 @@ def calculate_ylls(self, x: pd.DataFrame) -> float:
...
-            "results_dir": self.results_dir,
-            "random_seed": self.random_seed,
-            "input_draw": self.input_draw,
-        }
-        write_dataframe_to_parquet(
+        """Format dataframe and write out"""
+
+        results = results.reset_index()
+
+        if "cause_of_death" in results.columns:
```
rmudambi, line 159: Would it be better to have 
```suggestion
        if not self.aggregate_causes:
```

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -167,23 +149,50 @@ def calculate_ylls(self, x: pd.DataFrame) -> float:
...
+            results.rename(columns={"cause_of_death": COLUMNS.ENTITY}, inplace=True)
+        else:
+            # self.aggregate_causes is True
+            results[COLUMNS.ENTITY] = "all_causes"
+
+        results = results[results[COLUMNS.ENTITY] != "not_dead"]
+        results[COLUMNS.MEASURE] = measure
+
+        results.loc[results[COLUMNS.ENTITY] == "other_causes", COLUMNS.ENTITY_TYPE] = "cause"
```
rmudambi, line 168: You could replace this line and remove analogous the all_causes line below with 
```suggestion
        results[COLUMNS.ENTITY_TYPE] = "cause"
```

src/vivarium_public_health/metrics/mortality.py
```diff
@@ -167,23 +149,50 @@ def calculate_ylls(self, x: pd.DataFrame) -> float:
...
+            results[COLUMNS.ENTITY] == "other_causes", COLUMNS.SUB_ENTITY
+        ] = "other_causes"
+
+        results.loc[results[COLUMNS.ENTITY] == "all_causes", COLUMNS.ENTITY_TYPE] = "cause"
+        results.loc[
+            results[COLUMNS.ENTITY] == "all_causes", COLUMNS.SUB_ENTITY
+        ] = "all_causes"
+
+        for cause in self.causes_of_death:
+            results.loc[
```
rmudambi, line 179: Why not save out the mask to reduce duplication:
```python
cause_mask = results[COLUMNS.ENTITY] == cause.state_id
```

rmudambi (approved):

src/vivarium_public_health/metrics/mortality.py
stevebachmeier: It's use in two different methods though. I could define it once in `register_observations` and then partial that into `write_mortality_results` but that seems contrived.

src/vivarium_public_health/metrics/mortality.py
stevebachmeier: lol...I don't quite know why I did it this way. I even refer to `self.aggregate_causes` in the comment below

src/vivarium_public_health/metrics/mortality.py
stevebachmeier: I was initially doing it in one go via tuple but flake8 didn't like it for some reason so I split it into two lines like this and just didn't think about saving the mask

PR #429, opened 2024-05-31
Feature/sbachmei/mic 5095 update for refactored reporting

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5095

### Changes and notes
 
This updates the observers and related tests to use the latest approach
(namely, each observation has a `formatter` method that formats the
data but writing the results to disk is done by the simulation engine).

I also renamed the `metrics` folders to `results` for consistency with vivarium.

### Testing

tests are updated and all pass


src/vivarium_public_health/metrics/columns.py
```diff
@@ -0,0 +1,24 @@
+from typing import NamedTuple
+
+from vivarium.framework.results import VALUE_COLUMN
+
+
+class __Columns(NamedTuple):
```
stevebachmeier: This container was copied/pasted from the old/deleted `reporters.py` (once I removed the now-not-used `write_results` method all that was left was this object and figured it made more sense here.

rmudambi (approved):

PR #439, opened 2024-06-26
minor changes

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5127

### Changes and notes
 

- use the new `Observer.get_configuration_name()`
- exctract Mortality stratification categories into a causes_to_stratify attribute

### Testing

tests pass


albrja (approved):

src/vivarium_public_health/results/mortality.py
```diff
@@ -90,14 +90,18 @@ def columns_required(self) -> List[str]:
     # noinspection PyAttributeOutsideInit
     def setup(self, builder: Builder) -> None:
         self.clock = builder.time.clock()
-        self.config = builder.configuration.stratification.mortality
+        self.config = builder.configuration.stratification[self.get_configuration_name()]
```
rmudambi, line 93: Note: if you have rebased (I can't remember if you have), I recently introduced an instance method on `Component` called `get_configuration()`. You should use that here, but that can wait if you haven't rebased. If you're waiting, can you ticket it, so we don't forget to do it.

src/vivarium_public_health/results/observer.py
```diff
@@ -0,0 +1,34 @@
+from vivarium.framework.engine import Builder
+from vivarium.framework.results import Observer
+
+
+class HealthObserver(Observer):
```
rmudambi, line 5: Nit: I think `PublicHealthObserver` is better, since this is vivarium *public* health.

rmudambi: Also, I don't think these changes were supposed to be in this PR.

rmudambi (changes requested):

src/vivarium_public_health/results/mortality.py
stevebachmeier: Won't this not work b/c this is looking for "stratification" at the outer level?

rmudambi: You can override the `get_configuration()` method to do anything you want.

rmudambi (approved):

src/vivarium_public_health/results/mortality.py
stevebachmeier: Ah, yeah ok, I thought you thought it would just work out of the box

PR #442, opened 2024-07-01
Mic-5108/Bugfix/sort categories

albrja:
### Fixes bug where categories of risks were not being sorted correctly resulting in results that could not be reproduced.
- *Category*: Bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5108

### Changes and notes
-correctly sorts risk categories so results can be reproduced

### Testing
Ran two simulations back to back and output.hdf matched between both simulations.

### DIFF
--- a/src/vivarium_public_health/risks/distributions.py
+++ b/src/vivarium_public_health/risks/distributions.py
@@ -129,7 +129,7 @@ class EnsembleDistribution(RiskExposureDistribution):
             raise NotImplementedError("glnorm distribution is not supported")
         raw_weights = raw_weights[~glnorm_mask]
 
-        distributions = list(raw_weights["parameter"].unique())
+        distributions = sorted(list(raw_weights["parameter"].unique()))
 
         raw_weights = pivot_categorical(
             builder, self.risk, raw_weights, pivot_column="parameter", reset_index=False
@@ -287,7 +287,7 @@ class PolytomousDistribution(RiskExposureDistribution):
         self, exposure_data: Union[int, float, pd.DataFrame]
     ) -> Optional[List[str]]:
         if isinstance(exposure_data, pd.DataFrame):
-            return list(exposure_data["parameter"].unique())
+            return sorted(list(exposure_data["parameter"].unique()))
         return None
 
     def get_exposure_parameter_pipeline(self, builder: Builder) -> Pipeline:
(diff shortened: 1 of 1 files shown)


src/vivarium_public_health/risks/distributions.py
```diff
@@ -266,7 +266,7 @@ def ppf(self, quantiles: pd.Series) -> pd.Series:
 class PolytomousDistribution(RiskExposureDistribution):
     @property
     def categories(self) -> List[str]:
-        return self.lookup_tables["exposure"].value_columns
+        return sorted(self.lookup_tables["exposure"].value_columns)
```
hussain-jafari, line 269: Just to check - is it possible to sort this earlier?

albrja: I believe this is where we want to sort this because it is specific to this component but we could find a spot to check this in lookup table creation.

stevebachmeier: I still think the sorting should actually happen when/where you actually need it sorted, i.e. right at the `.cumsum`. That said, Rajan said just adding a comment is fine so I'll approve

stevebachmeier (approved):

patricktnast (approved):

src/vivarium_public_health/risks/distributions.py
rmudambi: I actually think the sorting should happen in `get_exposure_value_columns` in line 292, since that is where we originally lose the ordering, so if you looked at the LookupTable data in the debugger the columns would be ordered properly. That being said, it isn't a huge deal either way.

src/vivarium_public_health/risks/distributions.py
```diff
@@ -266,7 +266,9 @@ def ppf(self, quantiles: pd.Series) -> pd.Series:
 class PolytomousDistribution(RiskExposureDistribution):
     @property
     def categories(self) -> List[str]:
-        return self.lookup_tables["exposure"].value_columns
+        # These need to be sorted so the cumulative sum is in the ocrrect order of categories
```
rmudambi, line 269:
```suggestion
        # These need to be sorted so the cumulative sum is in the correct order of categories
```

PR #446, opened 2024-07-08
Bugfix/sbachmei/transition names to str

stevebachmeier:
### Description

- *Category*:  bugfix
- *JIRA issue*: na

### Changes and notes
 
We found that when using psimulate, we were getting a runtime error
when deserializing. It seems that the DiseaseObserver categories
(which were TransitionSTrings) were getting instantiated again
using their repr which causes an error - see https://jira.ihme.washington.edu/browse/MIC-5149

### Testing

Ran a small psimulate with the mnch maternal model

### DIFF
--- a/src/vivarium_public_health/results/__init__.py
+++ b/src/vivarium_public_health/results/__init__.py
@@ -1,3 +1,4 @@
+from .columns import COLUMNS
 from .disability import DisabilityObserver
 from .disease import DiseaseObserver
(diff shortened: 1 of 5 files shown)


hussain-jafari (approved):

hussain-jafari (changes requested):
failing tests

albrja (approved):

hussain-jafari (approved):

PR #451, opened 2024-07-24
build non log linear RR lookup table

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5161

### Changes and notes
Subclass risk effect to non log linear risk effect.
Initialize and define exposure column.
Build lookup table which 
 1) gets TMREL from artifact
 2) takes RR data from artifact, linearly interpolates over this data and evaluates at TMREL
 3) divide RR data from artifact by RR at TMREL and clips to be greater than 1
 3) returns exposure and RR points at the endpoints of the bin containing simulant exposure 

### Testing
Added diet high in sodium to child nutrition model repo and put in breakpoint to check that correct endpoints were returned given simulant exposure.

### DIFF
--- a/src/vivarium_public_health/risks/effect.py
+++ b/src/vivarium_public_health/risks/effect.py
@@ -13,6 +13,7 @@ from typing import Any, Callable, Dict, List, Tuple, Union
 
 import numpy as np
 import pandas as pd
+import scipy
 from layered_config_tree import ConfigurationError
 from vivarium import Component
 from vivarium.framework.engine import Builder
@@ -27,6 +28,8 @@ from vivarium_public_health.utilities import (
     TargetString,
     get_lookup_columns,
 )
+# TODO: discuss...
+from gbd_mapping import risk_factors
 
 
 class RiskEffect(Component):
@@ -329,7 +332,6 @@ class NonLogLinearRiskEffect(RiskEffect):
         return {
             self.name: {
                 "data_sources": {
-                    # TODO: replace with dataframe - possibly override elsewhere
                     "relative_risk": f"{self.risk}.relative_risk",
                     "population_attributable_fraction": f"{self.risk}.population_attributable_fraction",
                 },
@@ -364,19 +366,20 @@ class NonLogLinearRiskEffect(RiskEffect):
             configuration = self.configuration
 
         # get TMREL
-        # TODO: need access to risk TMRED
-        # TODO: can use entitykey or whatever we use in loader.py to get gbd_mapping entity
-        #if risk.tmred.distribution == 'uniform:
-        #    self.tmrel = np.random.uniform(risk.tmred.min, risk.tmred.max)
-        #elif risk.tmred.distribution == 'draws': # currently only for iron deficiency
-        #    raise DataAbornmalError('need to contact research team to get draws')
-        #else:
-        #    raise SomeError('no TMRED found for risk')
+        risk = risk_factors[self.risk.name]
+        if risk.tmred.distribution == 'uniform':
+            self.tmrel = np.random.uniform(risk.tmred.min, risk.tmred.max)
+        elif risk.tmred.distribution == 'draws': # currently only for iron deficiency
+            raise DataAbnormalError('TMRED has a non-uniform distribution. You will need to contact the research team to get this data.')
+        else:
+            raise ValueError(f'No TMRED found in gbd_mapping for risk {self.risk.name}')
 
         # calculate RR at TMREL
-        original_rrs = configuration.data_sources.relative_risk
+        rr_source = configuration.data_sources.relative_risk
+        original_rrs = self.get_filtered_data(builder, rr_source)
+        import pdb; pdb.set_trace()
         # TMREL for each age sex year row
-        demographic_cols = [col for col in original_rrs.index.names if col != 'parameter']
+        demographic_cols = [col for col in index_cols if col != 'parameter' and col != 'value']
         # for each combination of demographic cols
         # interpolate across exposures and RRs
         # raw_relative_risk_function = scipy.interpolate.interp1d(
--- a/src/vivarium_public_health/risks/effect.py
+++ b/src/vivarium_public_health/risks/effect.py
@@ -17,6 +17,8 @@ import scipy
 from layered_config_tree import ConfigurationError
 from vivarium import Component
 from vivarium.framework.engine import Builder
+from vivarium.framework.event import Event
+from vivarium.framework.population import SimulantData
 
 from vivarium_public_health.risks import Risk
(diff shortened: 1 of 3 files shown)


src/vivarium_public_health/risks/effect.py
```diff
@@ -316,3 +320,158 @@ def _get_risk_exposure_class(self, builder: Builder) -> Risk:
...
+        rr_data = rr_data.drop("parameter", axis=1)
+        rr_data[f"{self.risk.name}_exposure_start"] = rr_data["left_exposure"]
+        rr_data[f"{self.risk.name}_exposure_end"] = rr_data["right_exposure"]
+        # build lookup table
+        rr_value_cols = ["left_exposure", "left_rr", "right_exposure", "right_rr"]
+        self.lookup_tables["relative_risk"] = self.build_lookup_table(
+            builder, rr_data, rr_value_cols
+        )
+
+        # paf_data = self.get_filtered_data(
```
hussain-jafari, line 406: This was done temporarily to avoid having to put PAF data in the artifact.

src/vivarium_public_health/risks/effect.py
```diff
@@ -316,3 +320,158 @@ def _get_risk_exposure_class(self, builder: Builder) -> Risk:
...
+            .rename("rr_at_tmrel")
+        )
+        rr_data = original_rrs.merge(rrs_at_tmrel.reset_index())
+        rr_data["value"] = rr_data["value"] / rr_data["rr_at_tmrel"]
+        rr_data["value"] = np.clip(rr_data["value"], 1.0, np.inf)
+        rr_data = rr_data.drop("rr_at_tmrel", axis=1)
+
+        return rr_data
+
+    def get_target_modifier(
```
hussain-jafari, line 466: This function can be ignored because it will be updated in the next PR.

src/vivarium_public_health/risks/effect.py
```diff
@@ -316,3 +320,158 @@ def _get_risk_exposure_class(self, builder: Builder) -> Risk:
                 f"Risk effect model {self.name} requires a Risk component named {self.risk}"
             )
         return risk_exposure_component
+
+
+class NonLogLinearRiskEffect(RiskEffect):
+    """Risk effect for exposure-parametrized relative risks."""
```
hussain-jafari, line 326: This docstring will be updated in the final PR

src/vivarium_public_health/risks/effect.py
```diff
@@ -316,3 +320,158 @@ def _get_risk_exposure_class(self, builder: Builder) -> Risk:
...
+    # Lifecycle methods #
+    #####################
+
+    def on_initialize_simulants(self, pop_data: SimulantData) -> None:
+        exposure_col = pd.DataFrame(columns=self.columns_created)
+        self.population_view.update(exposure_col)
+
+    def on_time_step_prepare(self, event: Event) -> None:
+        # exposure_values = self.exposure(event.index)
+        # TODO: temporary workaround, investigate why exposure is always 0
```
albrja, line 361: I'm going to make a guess - exposure values are 0 because of some interaction that happens with subclassing RiskEffect. I'm curious what functionality you got from subclassing that component since you still had to write a lot of code and if it would just be worth inheriting from component instead.

albrja: FWIW I think you should subclass from RiskEffet but if this component is so different that you don't gain any value from it then maybe you don't need to

src/vivarium_public_health/risks/effect.py
albrja: You don't need data in the artifact though right? You can just configure it so that it loads it from that key.

src/vivarium_public_health/risks/effect.py
hussain-jafari: I've taken a look and I think there would be enough copy and paste that it's worth subclassing.

src/vivarium_public_health/risks/effect.py
```diff
@@ -316,3 +320,164 @@ def _get_risk_exposure_class(self, builder: Builder) -> Risk:
...
+    #################
+    # Setup methods #
+    #################
+
+    def build_all_lookup_tables(self, builder: Builder) -> None:
+        rr_data = self.get_relative_risk_data(builder)
+
+        # TODO: add check that rr_data is parametrize by exposure
+        def define_rr_intervals(df: pd.DataFrame) -> pd.DataFrame:
+            max_exposure_row = df.tail(1).copy()
```
hussain-jafari, line 377: add comment

src/vivarium_public_health/risks/effect.py
```diff
@@ -316,3 +320,164 @@ def _get_risk_exposure_class(self, builder: Builder) -> Risk:
...
+            max_exposure_row["parameter"] = 9999
+            rr_data = pd.concat([df, max_exposure_row]).reset_index()
+            rr_data["left_exposure"] = [0] + rr_data["parameter"][:-1].tolist()
+            rr_data["left_rr"] = [rr_data["value"].min()] + rr_data["value"][:-1].tolist()
```
stevebachmeier, line 381: Is this forcing the left_rr.min() at exposure 0? Is that the correct thing to do?

stevebachmeier (changes requested):
Just looking to make sure you confirmed your endpoints are correct

src/vivarium_public_health/risks/effect.py
```diff
@@ -316,3 +320,164 @@ def _get_risk_exposure_class(self, builder: Builder) -> Risk:
...
+
+        # calculate RR at TMREL
+        rr_source = configuration.data_sources.relative_risk
+        original_rrs = self.get_filtered_data(builder, rr_source)
+        demographic_cols = [
+            col for col in original_rrs.columns if col != "parameter" and col != "value"
+        ]
+
+        def get_rr_at_tmrel(rr_data: pd.DataFrame) -> float:
+            interpolated_rr_function = scipy.interpolate.interp1d(
```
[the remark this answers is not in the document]
hussain-jafari: It works fine in this case because we only need to return a single number, but down below we would need to return the actual interpolated function. I could have created a lookup table that returns a hexadecimal representation of the interpolated function for each demographic group and use that to evaluate the exposures of each sub-group but this way seems simpler.

src/vivarium_public_health/risks/effect.py
hussain-jafari: Good question - we do want to 1) fill all RR values below the lowest exposure to be the smallest relative risk which 2) will be 1 anyway because this exposure should be the TMREL.

stevebachmeier (approved):

PR #452, opened 2024-07-25
interpolate

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5201

### Changes and notes
Linearly interpolate RRs using lookup table output. 

### Testing
Spot checked a few RRs using simulant exposure values. Plotted interpolated results over original RR data and checked that the shapes were the same visually. A representative example is shown below:

interpolated_rrs
original_rrs

### DIFF
--- a/src/vivarium_public_health/risks/effect.py
+++ b/src/vivarium_public_health/risks/effect.py
@@ -371,6 +371,7 @@ class NonLogLinearRiskEffect(RiskEffect):
 
     def build_all_lookup_tables(self, builder: Builder) -> None:
         rr_data = self.get_relative_risk_data(builder)
+
         # TODO: add check that rr_data is parametrize by exposure
         def define_rr_intervals(df: pd.DataFrame) -> pd.DataFrame:
             max_exposure_row = df.tail(1).copy()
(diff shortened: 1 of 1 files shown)


src/vivarium_public_health/risks/effect.py
```diff
@@ -468,10 +469,15 @@ def get_target_modifier(
...
-            # use exposures and rr points to calculate m and b in y=mx+b
-            # ie m = (y2-y1)/(x2-x1) and b = y1 - mx1
-            # plug exposures into mx+b to get our RRs
+            exposure = self.population_view.get(index)[f"{self.risk.name}_exposure"]
+            x1, x2 = (
+                rr_intervals["left_exposure"].values,
+                rr_intervals["right_exposure"].values,
+            )
+            y1, y2 = rr_intervals["left_rr"].values, rr_intervals["right_rr"].values
+            m = (y2 - y1) / (x2 - x1)
```
stevebachmeier, line 478: Is it possible to have x2 = x1 and get a DivideByZero?

stevebachmeier (approved):

patricktnast:
>  Plotted interpolated results over original RR data and checked that the shapes were the same visually.

It would be awesome to post those plots here!

patricktnast (approved):

src/vivarium_public_health/risks/effect.py
hussain-jafari: Good question. x1 is the largest exposure for which an RR is defined that is less than or equal to a simulant's exposure and x2 is the smallest exposure for which an RR is defined that is greater than simulant's exposure, so they should never be equal.

hussain-jafari:
Good call @patricktnast I've attached some pictures in the description

PR #461, opened 2024-08-07
Feature/sbachmei/mic 5163 exclude unwanted results

stevebachmeier:
### Description

- *Category*:  feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5163

### Changes and notes

This implements the ability to exclude results via the model spec
with a new `stratification: excluded_categories` key. Key notes:
- SimpleCause is a new class to be used for causes that only
need a few attributes required by the observer (state_id, model,
and cause_type) as opposed to all the other goodies that come
with States.
- the "all_causes" disability pipline has been renamed from 
"disability_pipeline" to "all_causes.disability_pipeline"

### Testing

tests pass and I ran some small simulations against the maternal
and child models.

### DIFF
--- a/src/vivarium_public_health/results/disability.py
+++ b/src/vivarium_public_health/results/disability.py
@@ -77,8 +77,8 @@ class DisabilityObserver(PublicHealthObserver):
         registered stratifications and so cannot be excluded during the stratification
         call like other categories.
         """
-        causes_of_disability = list(
-            builder.components.get_components_by_type(tuple(self.disability_classes))
+        causes_of_disability = builder.components.get_components_by_type(
+            self.disability_classes
         )
         # Convert to SimpleCause instances and add on all_causes
         causes_of_disability = [
@@ -102,7 +102,7 @@ class DisabilityObserver(PublicHealthObserver):
 
         # Drop excluded causes
         if excluded_causes:
-            logger.info(
+            logger.debug(
                 f"'disability' has category exclusion requests: {excluded_causes}\n"
                 "Removing these from the allowable categories."
             )
--- a/src/vivarium_public_health/results/disability.py
+++ b/src/vivarium_public_health/results/disability.py
@@ -8,11 +8,12 @@ in the simulation.
 
 """
 
-from typing import Any, List, Type, Union
+from typing import Any, List, Union
 
 import pandas as pd
 from layered_config_tree import LayeredConfigTree
 from loguru import logger
+from pandas.api.types import CategoricalDtype
 from vivarium.framework.engine import Builder
 from vivarium.framework.values import Pipeline, list_combiner, union_post_processor
 
@@ -48,7 +49,7 @@ class DisabilityObserver(PublicHealthObserver):
     ##############
 
     @property
-    def disability_classes(self) -> List[Type]:
+    def disability_classes(self) -> list[type]:
         """The classes to be considered for causes of disability."""
         return [DiseaseState, RiskAttributableDisease]
 
@@ -71,16 +72,18 @@ class DisabilityObserver(PublicHealthObserver):
 
     def set_causes_of_disability(self, builder: Builder) -> None:
         """Set the causes of disability to be observed by removing any excluded
-        via the model spec from the list of all disability class causes.
+        via the model spec from the list of all disability class causes. We implement
+        exclusions here because disabilities are unique in that they are not
+        registered stratifications and so cannot be excluded during the stratification
+        call like other categories.
         """
         causes_of_disability = list(
             builder.components.get_components_by_type(tuple(self.disability_classes))
         )
         # Convert to SimpleCause instances and add on all_causes
-        all_causes = SimpleCause("all_causes", "all_causes", "cause")
         causes_of_disability = [
             SimpleCause.create_from_disease_state(cause) for cause in causes_of_disability
-        ] + [all_causes]
+        ] + [SimpleCause("all_causes", "all_causes", "cause")]
 
         excluded_causes = (
             builder.configuration.stratification.excluded_categories.to_dict().get(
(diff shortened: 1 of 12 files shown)


src/vivarium_public_health/results/mortality.py
```diff
@@ -90,28 +92,42 @@ def columns_required(self) -> List[str]:
...
+    def set_causes_of_death(self, builder: Builder) -> None:
+        causes_of_death = [
             cause
             for cause in builder.components.get_components_by_type(
                 tuple(self.mortality_classes)
             )
             if cause.has_excess_mortality
         ]
 
+        # Convert to SimpleCauses and add on other_causes and not_dead
```
albrja, line 106: Why do we have to do this now?

src/vivarium_public_health/results/simple_cause.py
```diff
@@ -0,0 +1,18 @@
...
+    ``SimpleCause`` instance.
+    """
+
+    state_id: str
+    model: str
+    cause_type: str
+
+    @classmethod
+    def create_from_disease_state(cls, disease_state):
+        return cls(disease_state.state_id, disease_state.model, disease_state.cause_type)
```
albrja, line 18: This is basically a class to store information since it doesn't have a __call__ method?

src/vivarium_public_health/results/simple_cause.py
```diff
@@ -0,0 +1,18 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class SimpleCause:
+    """A simple dataclass to represent the bare minimum information needed
+    for observers, e.g. 'all_causes' as a cause of disability. It also
+    includes a class method to coerce a provided disease state into a
```
rmudambi, line 8: Nit:
```suggestion
    includes a class method to convert a provided disease state into a
```

src/vivarium_public_health/results/simple_cause.py
```diff
@@ -0,0 +1,18 @@
...
+    ``SimpleCause`` instance.
+    """
+
+    state_id: str
+    model: str
+    cause_type: str
+
+    @classmethod
+    def create_from_disease_state(cls, disease_state):
```
rmudambi, line 17: Typing

tests/results/test_disease_observer.py
```diff
@@ -261,3 +272,59 @@ def test_different_results_per_disease(base_config, base_plugins):
...
+        "susceptible_to_human",
+        "turning",
+        "vampire",
+    }
+    transition_count_categories = {
+        "susceptible_to_human_to_turning",
+        "turning_to_vampire",
+    }
+    assert set(person_time["sub_entity"]) == person_time_categories - set(
+        transition_count_exclusions
```
rmudambi, line 326: Why is this using the `transition_count_exclusions`? Shouldn't it use the `person_time_exclusions`? Same question below.

rmudambi (approved):

src/vivarium_public_health/results/mortality.py
stevebachmeier: SimpleClass was added to easy in handling what is actually required for observers. An example is anemia in the maternal model, or "all_causes" for mortality. For anemia, we previously handled it by making a DiseaseState out of it and then setting it as a model, but that's hard to remember exactly what to do and also comes w/ a bunch of overhead for unused things (transitions, etc). All we _really_ need as far as observing the results of anemia are that it has state_id, model, and cause_type attributes. That's it. 

Probably more importantly, this class acts as a fail-quick device b/c if any given cause doesn't happen to have on of those attributes, it won't runtime error until the VERY end of the sim during results writing.

src/vivarium_public_health/results/simple_cause.py
stevebachmeier: see comment above

src/vivarium_public_health/results/disability.py
```diff
@@ -64,19 +68,54 @@ def __init__(self) -> None:
...
+
+    def set_causes_of_disability(self, builder: Builder) -> None:
+        """Set the causes of disability to be observed by removing any excluded
+        via the model spec from the list of all disability class causes. We implement
+        exclusions here because disabilities are unique in that they are not
+        registered stratifications and so cannot be excluded during the stratification
+        call like other categories.
+        """
+        causes_of_disability = list(
+            builder.components.get_components_by_type(tuple(self.disability_classes))
```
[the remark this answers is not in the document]
stevebachmeier: Hmm, this was my attempt to remove unnecessary list comprehension since whatever was returned could just be cast to `list`. But I think you're right, the actual return of `get_components_by_type` is already a list so no need to re-cast.

As for the tuple, it is indeed type-hinted that way. A quick look tells me it could take the `Union[Tuple, list]` but I'm hesitant to start down that rabbit hole unless you think it's worth it.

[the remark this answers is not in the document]
stevebachmeier: ok, was easy enough to allow for lists to be passed in and then coerce later to tuple. (tuple is needed down the line for an isinstance call)

src/vivarium_public_health/results/disability.py
```diff
@@ -64,19 +68,54 @@ def __init__(self) -> None:
...
+        excluded_causes = (
+            builder.configuration.stratification.excluded_categories.to_dict().get(
+                "disability", []
+            )
+        )
+
+        # Handle exclusions that don't exist in the list of causes
+        cause_names = [cause.state_id for cause in causes_of_disability]
+        unknown_exclusions = set(excluded_causes) - set(cause_names)
+        if len(unknown_exclusions) > 0:
```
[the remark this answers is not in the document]
stevebachmeier: But we use `unknown_exclusions` in the error message too. Would you prefer `if unknown_exclusions:` instead of a length check? (I prefer being explicit but don't feel strongly)

tests/results/test_disease_observer.py
stevebachmeier: Oof, scary - good catch

hussain-jafari (approved):
Approved expecting state and transition exclusions to be switched in the test suite :)

tests/results/test_disability_observer.py
```diff
@@ -230,3 +235,125 @@ def test_disability_accumulation(
...
+@pytest.mark.parametrize("exclusions", [[], ["all_causes"], ["all_causes", "sick_cause_0"]])
+def test_category_exclusions(
+    base_config,
+    base_plugins,
+    exclusions,
+):
+    """Integration test for the disability observer and the Results Management system."""
+
+    if len(exclusions) == 2:
+        pytest.skip(reason="FIXME: error when excluding all but one disability")
```
stevebachmeier, line 295: Not sure how high of a priority this is. Seems unlikely that we'd every exclude evertyhing but a single disability? Will make a separate ticket though b/c this PR has been open long enough.

stevebachmeier: rajan confirmed it does not have to be part of this PR

rmudambi (approved):

hussain-jafari (approved):

patricktnast (approved):

PR #463, opened 2024-08-08
non loglinear rrs

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5161

### Changes and notes
Add NonLogLinearRiskEffect.

### Testing
Added test and ran some minimal models with new component.

### DIFF
--- a/src/vivarium_public_health/risks/effect.py
+++ b/src/vivarium_public_health/risks/effect.py
@@ -26,11 +26,7 @@ from vivarium_public_health.risks.data_transformations import (
     pivot_categorical,
 )
 from vivarium_public_health.risks.distributions import MissingDataError
-from vivarium_public_health.utilities import (
-    EntityString,
-    TargetString,
-    get_lookup_columns,
-)
+from vivarium_public_health.utilities import EntityString, TargetString, get_lookup_columns
 
 
 class RiskEffect(Component):
--- a/src/vivarium_public_health/risks/effect.py
+++ b/src/vivarium_public_health/risks/effect.py
@@ -27,6 +27,7 @@ from vivarium_public_health.risks.data_transformations import (
 )
 from vivarium_public_health.risks.distributions import MissingDataError
 from vivarium_public_health.utilities import (
+    NUM_RR_EXPOSURE_VALUES,
     EntityString,
     TargetString,
     get_lookup_columns,
@@ -323,8 +324,19 @@ class RiskEffect(Component):
 
 
 class NonLogLinearRiskEffect(RiskEffect):
-    # TODO: update docstring
-    """Risk effect for exposure-parametrized relative risks."""
+    """A component to model the impact of an exposure-parametrized risk factor on
+    the target rate of some affected entity. This component will
+
+    1) read TMRED data from the artifact and define the TMREL
+    2) calculate the relative risk at TMREL by linearly interpolating over
+       relative risk data defined in the configuration
+    3) divide relative risk data from configuration by RR at TMREL
+       and clip to be greater than 1
+    4) build a LookupTable which returns the exposure and RR of the left and right edges
+       of the RR bin containing a simulant's exposure
+    5) use this LookupTable to modify the target pipeline by linearly interpolating
+       a simulant's RR value and multiplying it by the intended target rate
+    """
 
     ##############
     # Properties #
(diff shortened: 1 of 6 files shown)


src/vivarium_public_health/risks/effect.py
```diff
@@ -316,3 +321,185 @@ def _get_risk_exposure_class(self, builder: Builder) -> Risk:
...
+    #################
+    # Setup methods #
+    #################
+
+    def build_all_lookup_tables(self, builder: Builder) -> None:
+        rr_data = self.get_relative_risk_data(builder)
+
+        # check that rr_data has 1000 parameter values
+        exposure_values = rr_data["parameter"].values
+        if len(np.unique(exposure_values)) != NUM_RR_EXPOSURE_VALUES:
```
albrja, line 373: Is this constant being used other places? If it IS NOT then maybe this could be configurable.

src/vivarium_public_health/risks/effect.py
```diff
@@ -316,3 +321,185 @@ def _get_risk_exposure_class(self, builder: Builder) -> Risk:
...
+        def values_are_monotonically_increasing(df: pd.DataFrame) -> bool:
+            return np.all(df["parameter"].values[1:] >= df["parameter"].values[:-1])
+
+        group_is_increasing = rr_data.groupby(demographic_cols).apply(
+            values_are_monotonically_increasing
+        )
+        if not group_is_increasing.all():
+            raise ValueError(
+                "The parameter column in your relative risk data must be monotonically increasing to be used in NonLogLinearRiskEffect."
+            )
```
albrja, line 392: I'm assuming you are doing this because you care about the endpoint but I think a comment as to why this matters would be some helpful context.

src/vivarium_public_health/risks/effect.py
```diff
@@ -316,3 +321,185 @@ def _get_risk_exposure_class(self, builder: Builder) -> Risk:
...
+        self.lookup_tables["relative_risk"] = self.build_lookup_table(
+            builder, rr_data, rr_value_cols
+        )
+
+        paf_data = self.get_filtered_data(
+            builder, self.configuration.data_sources.population_attributable_fraction
+        )
+        self.lookup_tables["population_attributable_fraction"] = self.build_lookup_table(
+            builder, paf_data
+        )
```
albrja, line 433: Build all lookup tables method is really important especially when you are making a complicated component like this for the versatility it offers. It would be a lot easier to read if you made these local functions private class methods.. This would also make user's lives easier if they needed to subclass this component in the future.

albrja: You also did this with the relative risk data which I think you should rename to _get_relative_risk_data since that is only being used for this class I believe and there is a different method that gets the same data elsewhere (or at least there use to be before a bunch of this risk work).

src/vivarium_public_health/utilities.py
```diff
@@ -112,3 +112,6 @@ def get_lookup_columns(
         necessary_columns.remove("year")
 
     return list(necessary_columns)
+
+
+NUM_RR_EXPOSURE_VALUES = 1000
```
albrja, line 117: I stand by this should be configurable if it is only used for this new class but could have this as a default but this should be in a constants file and not utilities.

tests/risks/test_effect.py
```diff
@@ -398,3 +413,118 @@
...
+    rate = simulation.get_value("some_disease.incidence_rate")(
+        pop.index, skip_post_processor=True
+    )
+    expected_values = np.interp(
+        custom_exposure_values,
+        risk_effect_exposures,
+        np.array(risk_effect_rrs) / 2,  # RRs get divided by RR at TMREL
+    )
+
+    assert np.isclose(rate.values, expected_values, rtol=0.0000001).all()
```
albrja, line 530: I'd add a test or two that tests some of your errors you anticipated. I think it should be relatively easy to make a dataframe you want to throw a particular error and make a test component and initialize a simulation.

src/vivarium_public_health/risks/effect.py
```diff
@@ -316,3 +321,185 @@ def _get_risk_exposure_class(self, builder: Builder) -> Risk:
                 f"Risk effect model {self.name} requires a Risk component named {self.risk}"
             )
         return risk_exposure_component
+
+
+class NonLogLinearRiskEffect(RiskEffect):
+    """A component to model the impact of an exposure-parametrized risk factor on
```
patricktnast, line 327: this is a good docstring

patricktnast (approved):

src/vivarium_public_health/risks/effect.py
hussain-jafari: Honestly thinking about it now this seems unnecessarily restrictive. I was assuming all relative risk data for this component would come from GBD but I think its _possible_ that a researcher would want to define their own custom data. I see no harm in skipping this test and replacing it with checking for numeric data in the parameter column. That's also a better test because the component won't break if we have greater or fewer exposure values but if will if the data is not numeric.

src/vivarium_public_health/risks/effect.py
hussain-jafari: I split out the validation into its own function (which is also what happens in distributions), but I think this component is specific enough that I don't see it getting subclassed. Most likely get_relative_risk_data would need to change anyways, which is already overwriting a function inherited from the parent class.

src/vivarium_public_health/utilities.py
hussain-jafari: I wasn't sure where to put this but I saw other constants get imported from this file so I put it here. But it's getting removed anyway haha

tests/risks/test_effect.py
hussain-jafari: Good idea I'll add those.

PR #480, opened 2024-09-04
rr_source bugfix

hussain-jafari:
### Description
fix bug when defining rr_source in RiskEffect
- *Category*: bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5281

### Changes and notes
Only try getting relative risk data from a distribution when rr_source is a string. Otherwise (meaning rr_source is a float of dataframe) use get_filtered_data.
Move parameter from the index to columns when processing categorical data.
Add demographic dimensions key to mock artifact. 
Replace setup_risk_simulation with setup_risk_effect_simulation which is the same but uses a risk effect provided by the user instead of defaulting to NonLogLinearRiskEffect.
Add tests for when rr_source is a string, float, or DataFrame.

### Testing
Tests pass.

### DIFF
--- a/tests/risks/test_effect.py
+++ b/tests/risks/test_effect.py
@@ -458,17 +458,18 @@ class CustomExposureRisk(Component):
         return data
 
 
-def _setup_risk_simulation(
+def _setup_risk_effect_simulation(
     config: LayeredConfigTree,
     plugins: LayeredConfigTree,
     risk: Union[str, Risk],
+    risk_effect: RiskEffect,
     data: Dict[str, Any],
 ) -> InteractiveContext:
     components = [
         TestPopulation(),
         risk,
         SI("some_disease"),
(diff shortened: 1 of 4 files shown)


tests/risks/test_effect.py
```diff
@@ -413,6 +414,95 @@
 #                                                   source=lambda index: pd.Series(0.1, index=index))
 #
 #     assert np.allclose(from_yearly(0.1, time_step)*50, em(simulation.get_population().index))
+def _setup_risk_effect_simulation(
```
hussain-jafari, line 417: this is the same as the old _setup_risk_simulation except the user passes in a risk_effect that gets used as a component in the sim rather than defaulting to NonLogLinearRiskeffect

rmudambi (approved):
Looks good, but let's wait to release until we fix the bug referenced here: https://ihme.slack.com/archives/GCF5T9TDM/p1725466756408999

albrja (approved):

PR #487, opened 2024-10-28
Sbachmei/mic 5125/import subpackages to top level namespace

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5125

### Changes and notes
 This simply imports most of the classes to the top-level namespace
so that the user doesn't need to know the full filepaths.
Note:

- I only imported classes (not functions or constants) except for a few instances
where something else was already being imported at the sub-package level
- I did not include anything from the mslt subpackage
- I didi not include error classes

### Testing

tests pass; I confirmed a few imports

### DIFF
--- a/src/vivarium_public_health/__init__.py
+++ b/src/vivarium_public_health/__init__.py
@@ -47,6 +47,19 @@ from vivarium_public_health.results import (
     ResultsStratifier,
     SimpleCause,
 )
+from vivarium_public_health.risks import (
+    ContinuousDistribution,
+    DichotomousDistribution,
+    EnsembleDistribution,
+    LBWSGDistribution,
+    LBWSGRisk,
+    LBWSGRiskEffect,
+    NonLogLinearRiskEffect,
+    PolytomousDistribution,
+    Risk,
+    RiskEffect,
+    RiskExposureDistribution,
+)
 
 __all__ = [
     __author__,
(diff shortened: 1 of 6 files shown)


src/vivarium_public_health/__init__.py
```diff
@@ -28,6 +28,18 @@
...
+from vivarium_public_health.population import (
+    AgeOutSimulants,
+    BasePopulation,
+    FertilityAgeSpecificRates,
+    FertilityCrudeBirthRate,
+    FertilityDeterministic,
+    Mortality,
+    generate_population,
+)
+from vivarium_public_health.results import (
```
hussain-jafari, line 40: Should this be empty?

hussain-jafari (approved):

src/vivarium_public_health/__init__.py
stevebachmeier: Not sure what you mean. It's importing a bunch of things? lines 41-48. @hussain-jafari

hussain-jafari: Oh odd it looks empty when I look commit by commit but looking at the whole PR I see that it has a bunch of things being imported.

patricktnast (approved):

src/vivarium_public_health/__init__.py
stevebachmeier: Ah, I might have accidentally pushed an empty list in the wrong commit. Good eye!

rmudambi (commented):
I'm not sure we want everything available at the top level. I think we should have a conversation about this.

PR #489, opened 2024-11-01
Albrja/mic-5485/scaled-population

albrja:
### Add ScalePopulation component to population package.
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5485

### Changes and notes
-Adds component to scale base population

### Testing
Successfully ran simulation.

### DIFF
--- a/src/vivarium_public_health/population/__init__.py
+++ b/src/vivarium_public_health/population/__init__.py
@@ -3,5 +3,5 @@ from .add_new_birth_cohorts import (
     FertilityCrudeBirthRate,
     FertilityDeterministic,
 )
-from .base_population import BasePopulation, generate_population
+from .base_population import BasePopulation, ScaledPopulation, generate_population
 from .mortality import Mortality
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -207,10 +207,9 @@ class BasePopulation(Component):
 
 
 class ScaledPopulation(BasePopulation):
-    """Component for scaling the population structure based on a scaling factor.
-
-    This component is to be used in place of BasePopulation when all simulants are
-    a subset of the total population and need to be scaled up. The scaling factor
+    """This component is to be used in place of BasePopulation when all simulants are
+    a subset of the total population and need to be rescaled. The base population
+    structure is multiplied by a provided scaling factor. This scaling factor
     can be a dataframe passed in or a string that corresponds to an artifact key.
     If providing an artifact key, users can specify that in the configuration file.
     For example:
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -207,9 +207,26 @@ class BasePopulation(Component):
 
 
 class ScaledPopulation(BasePopulation):
-    """Component for scaling the population structure based on a scaling factor."""
+    """Component for scaling the population structure based on a scaling factor.
+    This component is to be used in place of the `BasePopulation` component when
+    all simulants in a simulation are a subset of a population and the user needs
+    to scale the population structure to account for the subset of the overall
+    population structure being treated as the total population within the simulation.
+    The scaling factor can be a dataframe passed in with python code or a string that
+    corresponds to an artifact key. If providing an artifact key, users can specify that
+    in the configuration file. For example:
 
-    def __init__(self, scaling_factor: str | float | pd.DataFrame):
+    .. code-block:: yaml
+
+    components:
+        vivarium_public_health:
+            population:
+                - ScaledPopulation("some.artifact.key")
+
+
+    """
+
+    def __init__(self, scaling_factor: str | pd.DataFrame):
         super().__init__()
         self.scaling_factor = scaling_factor
 
@@ -218,10 +235,10 @@ class ScaledPopulation(BasePopulation):
         population_structure = load_population_structure(builder)
         if isinstance(scaling_factor, pd.DataFrame):
             scaling_factor = scaling_factor.set_index(
-                [col for col in scaling_factor.columns if "value" not in col]
+                [col for col in scaling_factor.columns if col != "value"]
             )
             population_structure = population_structure.set_index(
-                [col for col in population_structure.columns if "value" not in col]
+                [col for col in population_structure.columns if col != "value"]
             )
         scaled_population_structure = (population_structure * scaling_factor).reset_index()
 
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -8,6 +8,8 @@ characteristics to simulants.
 
 """
 
+from __future__ import annotations
+
 from typing import Callable, Dict, Iterable, List
 
 import numpy as np
@@ -76,12 +78,11 @@ class BasePopulation(Component):
         # Validate configuration for deprecated keys
         self._validate_config_for_deprecated_keys()
 
-        source_population_structure = load_population_structure(builder)
+        source_population_structure = self._load_population_structure(builder)
         self.demographic_proportions = assign_demographic_proportions(
             source_population_structure,
             include_sex=self.config.include_sex,
         )
-
         self.randomness = self.get_randomness_streams(builder)
         self.register_simulants = builder.randomness.register_simulants
 
@@ -138,7 +139,6 @@ class BasePopulation(Component):
         demographic_proportions = self.get_demographic_proportions_for_creation_time(
             self.demographic_proportions, pop_data.creation_time.year
         )
-
         self.population_view.update(
             generate_population(
                 simulant_ids=pop_data.index,
@@ -201,6 +201,31 @@ class BasePopulation(Component):
                 f"Configuration key '{key}' will be deprecated in future versions of Vivarium "
                 f"Public Health. Use the new key '{mapper[key]}' instead."
             )
+
+    def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
+        return load_population_structure(builder)
+
+
+class ScaledPopulation(BasePopulation):
+    """Component for scaling the population structure based on a scaling factor."""
+
+    def __init__(self, scaling_factor: str | float | pd.DataFrame):
+        super().__init__()
+        self.scaling_factor = scaling_factor
+
+    def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
+        scaling_factor = self.get_data(builder, self.scaling_factor)
+        population_structure = load_population_structure(builder)
+        if isinstance(scaling_factor, pd.DataFrame):
+            scaling_factor = scaling_factor.set_index(
+                [col for col in scaling_factor.columns if "value" not in col]
+            )
+            population_structure = population_structure.set_index(
+                [col for col in population_structure.columns if "value" not in col]
+            )
+        scaled_population_structure = (population_structure * scaling_factor).reset_index()
+
+        return scaled_population_structure
 
 
 class AgeOutSimulants(Component):
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -208,6 +208,7 @@ class BasePopulation(Component):
 
 class ScaledPopulation(BasePopulation):
     """Component for scaling the population structure based on a scaling factor.
+
     This component is to be used in place of the `BasePopulation` component when
     all simulants in a simulation are a subset of a population and the user needs
     to scale the population structure to account for the subset of the overall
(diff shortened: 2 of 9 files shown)


src/vivarium_public_health/population/base_population.py
```diff
@@ -202,6 +202,31 @@ def _validate_config_for_deprecated_keys(self) -> None:
...
             )
 
+    def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
+        return load_population_structure(builder)
+
+
+class ScaledPopulation(BasePopulation):
+    """Component for scaling the population structure based on a scaling factor."""
+
+    def __init__(self, scaling_factor: str | float | pd.DataFrame):
```
stevebachmeier, line 212: Can scaling_factor be an int?

src/vivarium_public_health/population/base_population.py
```diff
@@ -202,6 +202,31 @@ def _validate_config_for_deprecated_keys(self) -> None:
...
+        super().__init__()
+        self.scaling_factor = scaling_factor
+
+    def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
+        scaling_factor = self.get_data(builder, self.scaling_factor)
```
stevebachmeier, line 217: Why this indirection? doesn't `get_data` just return the thing anyway unless it's a function?

src/vivarium_public_health/population/base_population.py
```diff
@@ -202,6 +202,31 @@ def _validate_config_for_deprecated_keys(self) -> None:
...
+        population_structure = load_population_structure(builder)
+        if isinstance(scaling_factor, pd.DataFrame):
+            scaling_factor = scaling_factor.set_index(
+                [col for col in scaling_factor.columns if "value" not in col]
```
stevebachmeier, line 221: nit `... if col != "value"`

src/vivarium_public_health/population/base_population.py
```diff
@@ -202,6 +202,31 @@ def _validate_config_for_deprecated_keys(self) -> None:
...
+            )
+            population_structure = population_structure.set_index(
+                [col for col in population_structure.columns if "value" not in col]
```
stevebachmeier, line 224: nit `... if col != "value"`

src/vivarium_public_health/population/base_population.py
```diff
@@ -202,6 +202,31 @@ def _validate_config_for_deprecated_keys(self) -> None:
...
+    def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
+        return load_population_structure(builder)
+
+
+class ScaledPopulation(BasePopulation):
+    """Component for scaling the population structure based on a scaling factor."""
+
+    def __init__(self, scaling_factor: str | float | pd.DataFrame):
+        super().__init__()
+        self.scaling_factor = scaling_factor
```
stevebachmeier, line 214: Should probably add docstring for scaling_factor and specifically call out that it is multiplicative

Should it be a property?

tests/population/test_base_population.py
```diff
@@ -283,6 +283,43 @@ def test__assign_demography_with_age_bounds_error(base_simulants, include_sex):
...
+    scalar_data["scalar"] = 1 - (scalar_data["index"] / len(scalar_data))
+    scalar_data["value"] = scalar_data["value"] * scalar_data["scalar"]
+    scalar_data.drop(columns=["scalar", "index", "age"], inplace=True)
+    scaled_pop = bp.ScaledPopulation(scalar_data)
+    sim2 = InteractiveContext(
+        components=[scaled_pop], configuration=config, plugin_configuration=base_plugins
+    )
+    scaled = sim2._data.load("population.structure")
+
+    assert (scaled["value"] >= unscaled["value"]).all()
```
stevebachmeier, line 320: Is it reasonable to actually assert that the ratios equal the scalar_data values on a row-by-row bases?

src/vivarium_public_health/population/__init__.py
rmudambi, line 1: We also want to add it to the top-level `__init__`

src/vivarium_public_health/population/base_population.py
rmudambi: I think we actually don't want to allow it to be a float or an int, since a scalar scaling factor would have no effect.

src/vivarium_public_health/population/base_population.py
```diff
@@ -202,6 +202,31 @@ def _validate_config_for_deprecated_keys(self) -> None:
                 f"Public Health. Use the new key '{mapper[key]}' instead."
             )
 
+    def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
+        return load_population_structure(builder)
+
+
+class ScaledPopulation(BasePopulation):
+    """Component for scaling the population structure based on a scaling factor."""
```
rmudambi, line 210: Can you expand on this docstring. What is this component for? How do you use it?

src/vivarium_public_health/population/base_population.py
rmudambi: The most likely use-case is this is an artifact key, in which case we do need to call `get_data()` to get the actual data from the artifact.

tests/population/test_base_population.py
rmudambi, line 320: This test isn't testing what you think it is. Both `scaled` and `unscaled` are just the data in the artifact at the key `"population.structure"`. They will be identical in both simulations, since both simulations are using the same artifact.

You want to test that the respective population components have demographic proportions that differ in the expected manner (i.e. `scaled_pop.demographic_proportions != base_pop.demographic_proportions`). Possibly you want to test the actual population generated by the scaled population approximates the scaled demographic proportions.

rmudambi (changes requested):

tests/population/test_base_population.py
albrja: @rmudambi I've updated this test however I can't compare the actual state tables because we have to mock the return value of `generate_population` so the state tables are the same. I personally think saying two dataframes not being equal isn't a good enough test so would love to hear if you have any other ideas.

src/vivarium_public_health/population/base_population.py
```diff
@@ -202,6 +202,48 @@ def _validate_config_for_deprecated_keys(self) -> None:
                 f"Public Health. Use the new key '{mapper[key]}' instead."
             )
 
+    def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
+        return load_population_structure(builder)
+
+
+class ScaledPopulation(BasePopulation):
+    """Component for scaling the population structure based on a scaling factor.
```
rmudambi, line 210: Nit: should have a new line after the first sentence.

src/vivarium_public_health/population/base_population.py
rmudambi: I don't think it needs to be a property, but I agree that the docs should make it clear that it is a a multiplicative scaling factor.

tests/population/test_base_population.py
```diff
@@ -283,6 +284,128 @@ def test__assign_demography_with_age_bounds_error(base_simulants, include_sex):
...
+        fuzzy_checker.fuzzy_assert_proportion(
+            observed_numerator=number_of_sims,
+            observed_denominator=len(pop.loc[pop["sex"] == sex]),
+            target_proportion=target_proportion,
+            name=f"scaled_pop_proportion_check_{sex}_{age_start}",
+        )
+
+
+def test_scaled_population_str_constructor(
+    config, base_plugins, mocker, fuzzy_checker: FuzzyChecker
```
albrja, line 347: This is the exact same test but the scalar data is put into the artifact and we access it by using the artifact key when creating the class instance.

rmudambi: If these are the same test, we should be parameterizing that test rather than making two separate tests. That should be simple to do.

tests/population/test_base_population.py
```diff
@@ -283,41 +284,126 @@ def test__assign_demography_with_age_bounds_error(base_simulants, include_sex):
...
+        number_of_sims = len(
+            pop.loc[
+                (pop["age"] >= age_start)
+                & (pop["age"] <= row_data["age_end"])
+                & (pop["sex"] == sex)
+            ]
+        )
+        fuzzy_checker.fuzzy_assert_proportion(
+            observed_numerator=number_of_sims,
+            observed_denominator=len(pop.loc[pop["sex"] == sex]),
```
rmudambi, line 340: I would have expected the observed denominator to be `len(pop)`, and the target proportion to be `row_data["value"] / pop["value"].sum()`. That makes this test much simpler and should be equivalent?

tests/test_utilities.py
```diff
@@ -116,3 +116,39 @@ def make_age_bins():
...
     return pd.DataFrame(index=idx).reset_index()
+
+
+def simple_pop_structure() -> pd.DataFrame:
+    # Create simple population structure
+    age_idx = pd.MultiIndex.from_tuples(
+        [
+            (0.0, 25.0, "Young People"),
+            (25.0, 50.0, "Old People"),
+            (50.0, 75.0, "Ancient People"),
```
rmudambi, line 127: Wow, these age group names are aggressive :P

rmudambi (changes requested):

src/vivarium_public_health/population/base_population.py
```diff
@@ -207,22 +207,43 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
...
 class ScaledPopulation(BasePopulation):
-    """Component for scaling the population structure based on a scaling factor."""
+    """Component for scaling the population structure based on a scaling factor.
 
-    def __init__(self, scaling_factor: str | float | pd.DataFrame):
+    This component is to be used in place of the `BasePopulation` component when
+    all simulants in a simulation are a subset of a population and the user needs
+    to scale the population structure to account for the subset of the overall
+    population structure being treated as the total population within the simulation.
+    The scaling factor can be a dataframe passed in with python code or a string that
```
stevebachmeier, line 216: nit: remove "passed in with python code"

src/vivarium_public_health/population/base_population.py
```diff
@@ -207,22 +207,43 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
 
 
 class ScaledPopulation(BasePopulation):
-    """Component for scaling the population structure based on a scaling factor."""
+    """Component for scaling the population structure based on a scaling factor.
 
-    def __init__(self, scaling_factor: str | float | pd.DataFrame):
+    This component is to be used in place of the `BasePopulation` component when
```
stevebachmeier, line 212: This is a loooooong sentence with a lot of the word "population". What about simply:

`This component is to be used in place of `BasePopulation` when all simulants are a subset of the total population and need to be scaled up.`

src/vivarium_public_health/population/base_population.py
```diff
@@ -207,22 +207,43 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
...
+        vivarium_public_health:
+            population:
+                - ScaledPopulation("some.artifact.key")
+
+
+    """
+
+    def __init__(self, scaling_factor: str | pd.DataFrame):
         super().__init__()
         self.scaling_factor = scaling_factor
```
stevebachmeier, line 232: Provide a docstring for the attr (triple-quoted and below the assignment). Mention the scaling factor is multiplicative

tests/test_utilities.py
stevebachmeier: Yeah. I think I must really be ancient b/c I view 25 year olds as babies.

tests/test_utilities.py
```diff
@@ -64,7 +64,7 @@ def make_uniform_pop_data(age_bin_midpoint=False):
     age_bins = [(b.age_start, b.age_end) for b in make_age_bins().itertuples()]
     sexes = ("Male", "Female")
     years = zip(range(1990, 2018), range(1991, 2019))
-    locations = (1, 2)
+    locations = list(range(1, 2))
```
stevebachmeier, line 67: This feels contrived. Are you aware `list(range(1, 2))` == [1]. Is this intended? If so, why not just [1]?

tests/test_utilities.py
```diff
@@ -116,3 +116,39 @@ def make_age_bins():
...
+        [
+            (0.0, 25.0, "Young People"),
+            (25.0, 50.0, "Old People"),
+            (50.0, 75.0, "Ancient People"),
+            (75.0, 100.0, "People Who Beat the Odds"),
+        ],
+        names=["age_start", "age_end", "age_group_name"],
+    )
+    age_df = pd.DataFrame(index=age_idx).reset_index()
+    age_bins = [(b.age_start, b.age_end) for b in age_df.itertuples()]
```
stevebachmeier, line 133: nit: use `bin` instead of `b`?

stevebachmeier (approved):

tests/population/test_add_new_birth_cohort.py
```diff
@@ -130,7 +130,7 @@ def test_FertilityCrudeBirthRate_extrapolate(base_config, base_plugins):
         }
     )
     pop_size = base_config.population.population_size
-    true_pop_size = 50000  # What's available in the mock artifact
+    true_pop_size = 25000  # What's available in the mock artifact
```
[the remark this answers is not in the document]
albrja: This is a downstream effect of fixing the bug in the artifact.

tests/population/test_base_population.py
```diff
@@ -283,41 +284,126 @@ def test__assign_demography_with_age_bounds_error(base_simulants, include_sex):
...
-    base_pop = bp.BasePopulation()
-    components = [base_pop]
+def test_scaled_population(config, base_plugins, mocker, fuzzy_checker: FuzzyChecker):
     config.update(
         {
             "population": {
-                "population_size": start_population_size,
-                "include_sex": include_sex,
+                "population_size": 1_000_000,
+                "include_sex": "Both",
```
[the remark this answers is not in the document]
albrja: No. I increased the sample size by a factor of 10.

tests/population/test_base_population.py
albrja: Updated the implementation of this.

tests/test_utilities.py
albrja: Yes - this was causing issues with population creation. It was creating a population that had both population 1 and 2 and the index for all sex age groups for both locations. But then when we create the population we were just setting location to 1 so we had duplicate index.

tests/population/test_base_population.py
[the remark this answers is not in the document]
rmudambi: I meant "Both"

rmudambi (approved):

PR #495, opened 2024-11-22
simplify initialization of DiseaseModel

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5568

### Changes and notes
-  Enable initializing a population of all newborns
- Refactor DiseaseModel to be able to leverage the `on_initialize_simulants`
  method of Machine

### Testing
- All tests pass
- Wrote some new tests

### DIFF
--- a/src/vivarium_public_health/disease/model.py
+++ b/src/vivarium_public_health/disease/model.py
@@ -10,15 +10,15 @@ transitions at simulation initialization and during transitions.
 """
 import warnings
 from collections.abc import Callable, Iterable
+from functools import partial
 from typing import Any
 
-import numpy as np
 import pandas as pd
 from vivarium.exceptions import VivariumError
 from vivarium.framework.engine import Builder
-from vivarium.framework.event import Event
 from vivarium.framework.population import SimulantData
 from vivarium.framework.state_machine import Machine
+from vivarium.types import LookupTableData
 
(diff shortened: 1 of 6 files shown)


src/vivarium_public_health/disease/model.py
```diff
@@ -120,12 +97,35 @@ def setup(self, builder: Builder) -> None:
...
+            initialization_table_name = "birth_prevalence"
+        else:
+            initialization_table_name = "prevalence"
+
+        for state in self.states:
+            state.lookup_tables["initialization_weights"] = state.lookup_tables[
+                initialization_table_name
+            ]
+
+        super().on_initialize_simulants(pop_data)
```
albrja, line 128: Whats the reason for doing this at the end of the method?

albrja: It seems like it doesn't matter

rmudambi: The reason this matters is that `BaseDiseaseState` and its subclasses don't set their initialization weights when they're created. They only create `prevalence` and `birth_prevalence` lookup tables. `Machine` uses the `initialization_weights` lookup table, so we need to set that lookup table appropriately. If `age_end` is 0 we should use birth prevalence and otherwise, prevalence.

hussain-jafari (approved):

patricktnast (approved):

PR #501, opened 2024-12-17
Albrja/mic 5539/bugfix/population-age-bounds

albrja:
### Bugfix so simulants are initialized within the configured age bounds
- *Category*: Bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5539

### Changes and notes
-fix bug where simulants were getting initialized outside of the configured age bounds

### DIFF
--- a/src/vivarium_public_health/population/data_transformations.py
+++ b/src/vivarium_public_health/population/data_transformations.py
@@ -111,7 +111,6 @@ def rescale_binned_proportions(
         raise ValueError(
             "Provided population data is insufficient to model the requested age range."
         )
-
     age_start = max(pop_data["age_start"].min(), age_start)
     age_end = min(pop_data["age_end"].max(), age_end) - 1e-8
     pop_data = _add_edge_age_groups(pop_data.copy())
@@ -126,7 +125,6 @@ def rescale_binned_proportions(
             (sub_pop["age_start"] <= age_start) & (age_start < sub_pop["age_end"])
         ]
         padding_bin = sub_pop[sub_pop["age_end"] == float(min_bin["age_start"].iloc[0])]
-
         min_scale = (float(min_bin["age_end"].iloc[0]) - age_start) / float(
             min_bin["age_end"].iloc[0] - min_bin["age_start"].iloc[0]
         )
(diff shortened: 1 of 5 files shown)


tests/population/test_base_population.py
```diff
@@ -378,3 +378,26 @@ def _check_locations(simulants):
             1 / len(simulants.location.unique()),
             abs_tol=0.01,
         )
+
+
+@pytest.mark.parametrize("age_start, age_end", [(0, 100), (2, 52), (21, 67), (12, 48)])
```
stevebachmeier, line 383: Might be worth adding a (0, 1) parameter?

stevebachmeier, line 383: I pulled this branch to try and understand the new logic in data_tranformations.py and found that this test passes even when I commented that new logic out. i.e. we need a test that fails prior to the logic but passes with it

src/vivarium_public_health/population/data_transformations.py
```diff
@@ -165,6 +163,21 @@ def rescale_binned_proportions(
...
+        )
+        pop_data.loc[padding_bin.index, "age"] = float(
+            (
+                pop_data.loc[padding_bin.index, "age_start"].iloc[0]
+                + pop_data.loc[padding_bin.index, "age_end"].iloc[0]
+            )
+            / 2
+        )
 
     return pop_data[col_order].sort_values(_SORT_ORDER).reset_index(drop=True)
```
stevebachmeier, line 182: I know this isn't specific to this PR, but it looks like whatever is going on in this method has deviated from the docstring. i.e. the docstring says it returns "Table with the same columns as `pop_data` where all bins outside the range (age_start, age_end) have been discarded", but it looks like it still returns rows outside the range as well.

I feel like there's a bug floating around in here somewhere...

src/vivarium_public_health/population/data_transformations.py
```diff
@@ -165,6 +163,21 @@ def rescale_binned_proportions(
 
         pop_data.loc[max_bin.index, "age_end"] = age_end
         pop_data.loc[padding_bin.index, "age_start"] = age_end
+        # Update age to be within bounds
```
stevebachmeier, line 166: I don't understand what this logic is doing (or really, what any of the logic in this function is doing), but the returns seem weird to me. Let's talk.

stevebachmeier (changes requested):

src/vivarium_public_health/population/data_transformations.py
stevebachmeier: At the very least it doesn't seem to be updating both sexes the same:

image

PR #514, opened 2025-03-18
Albrja/mic-5451/disease-components

albrja:
### Update required resources for disease components
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5451

### Changes and notes
-update required resources

### DIFF
--- a/src/vivarium_public_health/disease/special_disease.py
+++ b/src/vivarium_public_health/disease/special_disease.py
@@ -16,6 +16,7 @@ import pandas as pd
 from vivarium import Component
 from vivarium.framework.event import Event
 from vivarium.framework.population import SimulantData
+from vivarium.framework.resource import Resource
 from vivarium.framework.values import list_combiner, union_post_processor
 
 from vivarium_public_health.disease.transition import TransitionString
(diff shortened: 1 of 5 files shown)


stevebachmeier:
Can you remove James from the CODEOWNERS?

src/vivarium_public_health/disease/special_disease.py
```diff
@@ -171,16 +168,20 @@ def setup(self, builder):
         self.disability_weight = builder.value.register_value_producer(
             f"{self.cause.name}.disability_weight",
             source=self.compute_disability_weight,
+            component=self,
             requires_columns=get_lookup_columns(
```
stevebachmeier, line 172: Isn't part of this ticket to also get rid of all of these deprecated args and replace w/ `required_resources`?

src/vivarium_public_health/disease/state.py
```diff
@@ -465,19 +462,22 @@ def setup(self, builder: Builder) -> None:
...
 
         builder.value.register_value_modifier(
             "mortality_rate",
             modifier=self.adjust_mortality_rate,
-            requires_values=[self.excess_mortality_rate_pipeline_name],
+            component=self,
+            required_resources=[self.excess_mortality_rate],
         )
 
         self.randomness_prevalence = self.get_randomness_prevalence(builder)
```
rmudambi, line 483: This was previously not being done, but technically, `DiseaseState` should add `self.randomness_prevalence` to its `initialization_requirements`. So you should override that property here like:
```python
def initialization_requirements(self) -> list[str | Resource]:
    return super().initialization_requirements() + [self.randomness_prevalence]
```

rmudambi (approved):

stevebachmeier (approved):

PR #521, opened 2025-05-14
Albrja/mic-6068/rate-to-probability-configuration

albrja:
### Update RateTransition to have rate conversion type configurable
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6068

### Changes and notes
-Update RateTransition to have rate conversion type configurable

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+**4.1.0 - 05/14/25**
+
+  - Feature: Update RateTransition to have configuration key for rate conversion type
+
 **4.0.2 - 05/01/25**
 
   - Bugfix: Fix configuration key for untracking age in FertilityCrudeBirthRate
--- a/src/vivarium_public_health/disease/transition.py
+++ b/src/vivarium_public_health/disease/transition.py
@@ -48,6 +48,7 @@ class RateTransition(Transition):
                 "data_sources": {
                     "transition_rate": self._rate_source,
                 },
+                "rate_conversion_type": "linear",
             },
         }
 
@@ -184,7 +185,11 @@ class RateTransition(Transition):
     ##################
 
     def _probability(self, index: pd.Index) -> pd.Series:
-        return pd.Series(rate_to_probability(self.transition_rate(index)))
+        return pd.Series(rate_to_probability(
+            self.transition_rate(index),
+            rate_conversion_type=self.configuration["rate_conversion_type"],
+            )
+        )
 
 
 class ProportionTransition(Transition):


CHANGELOG.rst
```diff
@@ -1,3 +1,7 @@
+**4.1.0 - 05/14/25**
```
albrja, line 1: This is a new feature and dependent upon the vivarium update but I would also be fine with a patch release

rmudambi (commented):
Looks good, but we need a test

stevebachmeier (approved):

tests/disease/test_disease.py
```diff
@@ -554,3 +557,99 @@ def test_artifact_transition_keys(mocker, disease):
...
+    with patch(
+        "vivarium_public_health.disease.transition.rate_to_probability", return_value=1.0
+    ) as mock_rate_to_probability:
+        idx = pd.Index(list(range(10)))
+        transition._probability(idx)
+        calls = mock_rate_to_probability.call_args_list
+        for call_args in calls:
+            arg1, arg2 = call_args[0][0], call_args[1]["rate_conversion_type"]
+            assert arg1.index.equals(idx)
+            assert arg2 == rate_conversion_type
```
rmudambi, line 609: Does this not work?
```python
mock_rate_to_probability.assert_called_once_with(idx, rate_conversion_type)
```

tests/disease/test_disease.py
```diff
@@ -554,3 +557,99 @@ def test_artifact_transition_keys(mocker, disease):
...
+        get_data_functions={"incidence_rate": lambda builder, _: builder.data.load(key)},
+    )
+    healthy.transition_set.append(transition)
+    sick.transition_set.append(another_transition)
+    model = DiseaseModel(disease, initial_state=healthy, states=[healthy, sick])
+
+    base_config.update(
+        {
+            f"{transition.name}": {
+                "rate_conversion_type": "invalid",
```
rmudambi, line 640: You should set this to "linear", because we don't want to get fooled into thinking this is failing because "invalid" is not a real conversion type - it's failing because it doesn't match "exponential" below.

tests/disease/test_disease.py
albrja: No because they take different args but I should be able to get it to work like this

albrja: Im going to see if there is a way to do that with a series

patricktnast (approved):

tests/disease/test_disease.py
albrja: Updated

rmudambi (approved):

PR #523, opened 2025-05-23
Albrja/mic-5621/observer-get-configuration

albrja:
### Overrides get_configuration in observers to use super class method
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5621

### Changes and notes
-override get_configuration for observer classes
-DOES NOT update method for disease and risk observers due to their configuration via self.cause and self.risk and not the observer component name


rmudambi (approved):
Love all the removed boilerplate!

PR #528, opened 2025-07-10
Fix bug in validating crude birth rates

albrja:
### Fix bug validating population configuration for potetntially non-existent key
- *Category*: Bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-XYZ

### DIFF
--- a/src/vivarium_public_health/population/data_transformations.py
+++ b/src/vivarium_public_health/population/data_transformations.py
@@ -577,8 +577,10 @@ def rescale_final_age_bin(builder, population_data):
 
 
 def validate_crude_birth_rate_data(builder, data_year_max):
-    untracking_age = builder.configuration.population.to_dict().get("untracking_age", None)
-    if untracking_age and builder.configuration.population.age_end != untracking_age:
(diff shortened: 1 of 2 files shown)


src/vivarium_public_health/population/data_transformations.py
```diff
@@ -577,8 +577,10 @@ def rescale_final_age_bin(builder, population_data):
 
 
 def validate_crude_birth_rate_data(builder, data_year_max):
-    untracking_age = builder.configuration.population.to_dict().get("untracking_age", None)
-    if untracking_age and builder.configuration.population.age_end != untracking_age:
+    population_config = builder.configuration.population.to_dict()
+    untracking_age = population_config.get("untracking_age", None)
+    age_end = population_config.get("age_end", None)
+    if untracking_age and age_end and age_end != untracking_age:
```
albrja, line 583: We want to use untrackign age but if age_end also exists they must be equal. They issue is we do not require age_end to be a key anymore.

patricktnast (approved):

PR #539, opened 2025-08-25
Albrja/mic 6322/update scaled population

albrja:
### Make ScaledPopulation able to handle data inputs with multiple years.
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6322

### Changes and notes
-adds data transformation function to ScaledPopulation component to format population_structure and scaling_factor
-allows for data inputs to have multiple years

### DIFF
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -266,7 +266,9 @@ class ScaledPopulation(BasePopulation):
             set(scaling_factor.index.get_level_values("year_start"))
         )
         scale_year_index = np.digitize(year, scale_reference_years).item() - 1
-        # Subset to start year of simulation or closest year
+        # Subset to start year of simulation or earliest year. E.g. if start year = 2021 and pop
+        # structure has 2021, we will subset to 2021. If pop structure minimum year is 2025, we
+        # will subset to 2025.
         population_structure = population_structure.loc[
             population_structure.index.get_level_values("year_start")
             == pop_reference_years[pop_year_index]
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -235,23 +235,18 @@ class ScaledPopulation(BasePopulation):
             raise ValueError(
                 f"Scaling factor must be a pandas DataFrame. Provided value: {scaling_factor}"
             )
+        start_year = builder.configuration.time.start.year
         population_structure, scaling_factor = self._format_data_inputs(
-            population_structure, scaling_factor
+            population_structure, scaling_factor, start_year
         )
 
         return (population_structure * scaling_factor).reset_index()
 
     def _format_data_inputs(
-        self, pop_structure: pd.DataFrame, scalar_data: pd.DataFrame
+        self, pop_structure: pd.DataFrame, scalar_data: pd.DataFrame, year: int
     ) -> tuple[pd.DataFrame, pd.DataFrame]:
         """Data cleaning function to check whether scalar_data and population structure are compatible for scaling
-        the population structure of a simulation. This method will do any necessary transformations to
-        deal with any of the following use cases:
-        1. Population structure and scaling factor each have one year of data.
-        2. Population structure has multiple years of data, and scaling factor has one year of data.
-        3. Both population structure and scaling factor have multiple years of data.
-        4. Population structure has one year of data, and scaling factor has multiple years of data in which
-        case a ValueError will be raised."""
+        the population structure of a simulation."""
 
         scaling_factor = scalar_data.set_index(
             [col for col in scalar_data.columns if col != "value"]
@@ -262,24 +257,24 @@ class ScaledPopulation(BasePopulation):
         if "year_start" not in scaling_factor.index.names:
             return population_structure, scaling_factor
 
-        pop_years = set(population_structure.index.get_level_values("year_start"))
-        scale_years = set(scaling_factor.index.get_level_values("year_start"))
-        if len(pop_years) > 1:
-            if len(scale_years) > 1:
-                if not pop_years == scale_years:
-                    raise ValueError(
-                        f"Population years {pop_years} and scaling years {scale_years} must match if both "
-                        "multiple years of data."
-                    )
-            if len(scale_years) == 1:
-                scaling_factor.index = scaling_factor.index.droplevel(
-                    ["year_start", "year_end"]
-                )
-        if len(pop_years) == 1 and len(scale_years) > 1:
-            raise ValueError(
-                f"Population years {pop_years} must be the same as scaling years {scale_years} or scaling years "
-                "must be a subset of population years."
-            )
+        # Subset to start year of simulation or closest year
+        pop_reference_years = sorted(
+            set(population_structure.index.get_level_values("year_start"))
+        )
+        pop_year_index = np.digitize(year, pop_reference_years).item() - 1
+        scale_reference_years = sorted(
+            set(scaling_factor.index.get_level_values("year_start"))
+        )
+        scale_year_index = np.digitize(year, scale_reference_years).item() - 1
+        # Subset to start year of simulation or closest year
+        population_structure = population_structure.loc[
+            population_structure.index.get_level_values("year_start")
+            == pop_reference_years[pop_year_index]
+        ]
+        scaling_factor = scaling_factor.loc[
+            scaling_factor.index.get_level_values("year_start")
+            == scale_reference_years[scale_year_index]
+        ]
 
         return population_structure, scaling_factor
 
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -235,15 +235,50 @@ class ScaledPopulation(BasePopulation):
             raise ValueError(
                 f"Scaling factor must be a pandas DataFrame. Provided value: {scaling_factor}"
             )
-        scaling_factor = scaling_factor.set_index(
-            [col for col in scaling_factor.columns if col != "value"]
+        formatted_pop_structure, formatted_scaling_factor = self._format_data_inputs(
+            population_structure, scaling_factor
         )
-        population_structure = population_structure.set_index(
-            [col for col in population_structure.columns if col != "value"]
-        )
-        scaled_population_structure = (population_structure * scaling_factor).reset_index()
+
+        scaled_population_structure = (
+            formatted_pop_structure * formatted_scaling_factor
+        ).reset_index()
 
         return scaled_population_structure
+
+    def _format_data_inputs(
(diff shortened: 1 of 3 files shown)


src/vivarium_public_health/population/base_population.py
```diff
@@ -235,16 +235,56 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
...
+                if not pop_years == scale_years:
+                    raise ValueError(
+                        f"Population years {pop_years} and scaling years {scale_years} must match if both "
+                        "multiple years of data."
+                    )
+            if len(scale_years) == 1:
+                scaling_factor.index = scaling_factor.index.droplevel(
+                    ["year_start", "year_end"]
+                )
+        if len(pop_years) == 1 and len(scale_years) > 1:
```
albrja, line 280: The idea here is I am not sure why the artifact would have multiple years for scaling factor but not population and what year would we want to use to scale the population structure. We could do the reverse implementation of above where we drop year index levels from population and multiply but that didn't seem quite right to me.

tests/population/test_base_population.py
```diff
@@ -301,7 +303,7 @@ def test_scaled_population(
 
     # Simple pop data
     pop_structure = simple_pop_structure()
-    # Simple scalar data to pass to ScaledPopulation
+    # Simple scalar data to pass to pulation
```
stevebachmeier, line 306: typo

src/vivarium_public_health/population/base_population.py
```diff
@@ -235,16 +235,56 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
...
-            [col for col in scaling_factor.columns if col != "value"]
+        formatted_pop_structure, formatted_scaling_factor = self._format_data_inputs(
+            population_structure, scaling_factor
         )
-        population_structure = population_structure.set_index(
-            [col for col in population_structure.columns if col != "value"]
-        )
-        scaled_population_structure = (population_structure * scaling_factor).reset_index()
+
+        scaled_population_structure = (
```
stevebachmeier, line 242: nit: just return this (though I think we disagree here sometimes...I don't feel that strongly!)

src/vivarium_public_health/population/base_population.py
```diff
@@ -235,16 +235,56 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
             raise ValueError(
                 f"Scaling factor must be a pandas DataFrame. Provided value: {scaling_factor}"
             )
-        scaling_factor = scaling_factor.set_index(
-            [col for col in scaling_factor.columns if col != "value"]
+        formatted_pop_structure, formatted_scaling_factor = self._format_data_inputs(
```
stevebachmeier, line 238: If you want to save typing and needing to think about formatted vs non-formatted variables, you can just actually return anything. You didn't make copies in `_format_data_inputs` so you're really just changing the original dataframes anyway and "mirroring" them with the new variables

src/vivarium_public_health/population/base_population.py
```diff
@@ -235,16 +235,56 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
...
+        self, pop_structure: pd.DataFrame, scalar_data: pd.DataFrame
+    ) -> tuple[pd.DataFrame, pd.DataFrame]:
+        """Data cleaning function to check whether scalar_data and population structure are compatible for scaling
+        the population structure of a simulation. This method will do any necessary transformations to
+        deal with two use cases where either population structure has multiple years of data, or both data
+        inputs have multiple years of data."""
+
+        scaling_factor = scalar_data.set_index(
+            [col for col in scalar_data.columns if col != "value"],
+            drop=True,
```
stevebachmeier, line 258: nit: drop=True is the default

src/vivarium_public_health/population/base_population.py
```diff
@@ -235,16 +235,56 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
...
+            formatted_pop_structure * formatted_scaling_factor
+        ).reset_index()
 
         return scaled_population_structure
 
+    def _format_data_inputs(
+        self, pop_structure: pd.DataFrame, scalar_data: pd.DataFrame
+    ) -> tuple[pd.DataFrame, pd.DataFrame]:
+        """Data cleaning function to check whether scalar_data and population structure are compatible for scaling
+        the population structure of a simulation. This method will do any necessary transformations to
```
stevebachmeier, line 252: Can you expand this a bit? I really don't understand why this function exists based on this docstring.

src/vivarium_public_health/population/base_population.py
```diff
@@ -235,16 +235,56 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
...
+
+        scaling_factor = scalar_data.set_index(
+            [col for col in scalar_data.columns if col != "value"],
+            drop=True,
+        )
+        population_structure = pop_structure.set_index(
+            [col for col in pop_structure.columns if col != "value"],
+            drop=True,
+        )
+        if "year_start" not in scaling_factor.index.names:
```
stevebachmeier, line 264: so this means that the scaling is to be applied to all years?

src/vivarium_public_health/population/base_population.py
```diff
@@ -235,16 +235,51 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
...
+            return population_structure, scaling_factor
+
+        pop_years = set(population_structure.index.get_level_values("year_start"))
+        scale_years = set(scaling_factor.index.get_level_values("year_start"))
+        if len(pop_years) > 1:
+            if len(scale_years) > 1:
+                if not pop_years == scale_years:
```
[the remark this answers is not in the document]
stevebachmeier: Why? Is it possible that scale_years has the entirety of pop_years plus extras?

src/vivarium_public_health/population/base_population.py
```diff
@@ -235,16 +235,56 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
...
+            [col for col in pop_structure.columns if col != "value"],
+            drop=True,
+        )
+        if "year_start" not in scaling_factor.index.names:
+            return population_structure, scaling_factor
+
+        pop_years = set(population_structure.index.get_level_values("year_start"))
+        scale_years = set(scaling_factor.index.get_level_values("year_start"))
+        if len(pop_years) > 1:
+            if len(scale_years) > 1:
```
stevebachmeier, line 270: combine this: `if len(scale_years) > 1 and pop_years != scale_years:`

src/vivarium_public_health/population/base_population.py
```diff
@@ -235,16 +235,56 @@ def _load_population_structure(self, builder: Builder) -> pd.DataFrame:
...
+                if not pop_years == scale_years:
+                    raise ValueError(
+                        f"Population years {pop_years} and scaling years {scale_years} must match if both "
+                        "multiple years of data."
+                    )
+            if len(scale_years) == 1:
```
stevebachmeier, line 276: This feels weird. so the population has multiple years and scale years has a single year. This is effectively using that one year of scaling values across all population years, right? Is that really correct?

If that _is_ correct, I'd recommend moving this logic up near where you returned them before when it didn't have years since this is kinda doing the same thing

src/vivarium_public_health/population/base_population.py
stevebachmeier: Kinda like I said above, is it really not ever going to happen where the scaling might have multiple years and as long as one of them matches the population then you're good?

src/vivarium_public_health/population/base_population.py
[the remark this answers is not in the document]
albrja: Response will be in other similar comment you have.

src/vivarium_public_health/population/base_population.py
albrja: Yes if the year index levels aren't in scaling factor it will be able to be cast across all years in the population structure.

tests/population/test_base_population.py
```diff
@@ -349,6 +351,117 @@ def test_scaled_population(
...
+        # Both have multiple years - should be able to multiply together directly
+        scalar_data = pd.DataFrame(
+            {
+                "age_start": [0.0, 0.0, 25.0, 25.0, 50.0, 50.0, 75.0, 75.0] * 3,
+                "age_end": [25.0, 25.0, 50.0, 50.0, 75.0, 75.0, 100.0, 100.0] * 3,
+                "sex": ["Male", "Female"] * 12,
+                "year_start": [2021] * 8 + [2022] * 8 + [2023] * 8,
+                "year_end": [2022] * 8 + [2023] * 8 + [2024] * 8,
+                "value": [
+                    1.0,
```
stevebachmeier, line 425: Do these really need to all be random different values? Can you just use the same value, e.g. `"value": [1.1]*24`?

rmudambi (approved):

src/vivarium_public_health/population/base_population.py
```diff
@@ -262,24 +257,24 @@ def _format_data_inputs(
...
-            )
+        # Subset to start year of simulation or closest year
+        pop_reference_years = sorted(
+            set(population_structure.index.get_level_values("year_start"))
+        )
+        pop_year_index = np.digitize(year, pop_reference_years).item() - 1
+        scale_reference_years = sorted(
+            set(scaling_factor.index.get_level_values("year_start"))
+        )
+        scale_year_index = np.digitize(year, scale_reference_years).item() - 1
```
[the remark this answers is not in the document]
stevebachmeier: How about 
```
diffs = [abs(x - year) for x in pop_reference_years]
pop_reference_years[diffs.index(min(diffs))]
```

or if you want to get fancy: `min(pop_reference_years, key=lambda x: abs(x - year))`

This is almost certainly slower than np.digitize but is more readable imo and is properly rounding.

```
pop_reference_years = [2025, 2026, 9999]


>>> diffs = [abs(x - 2025) for x in pop_reference_years]
>>> pop_reference_years[diffs.index(min(diffs))]
2025
>>> diffs = [abs(x - 2026) for x in pop_reference_years]
>>> pop_reference_years[diffs.index(min(diffs))]
2026
>>> diffs = [abs(x - 9999) for x in pop_reference_years]
>>> pop_reference_years[diffs.index(min(diffs))]
9999

>>> diffs = [abs(x - 9998) for x in pop_reference_years]
>>> pop_reference_years[diffs.index(min(diffs))]
9999

>>> diffs = [abs(x - 10000) for x in pop_reference_years]
>>> pop_reference_years[diffs.index(min(diffs))]
9999

>>> diffs = [abs(x - 1) for x in pop_reference_years]
>>> pop_reference_years[diffs.index(min(diffs))]
2025
```

stevebachmeier (approved):

PR #540, opened 2025-09-09
Raise if bin index is out of range

stevebachmeier:
### Description

- *Category*:  bugfix
- *JIRA issue*: na

### Changes and notes
 

np.digitize - 1 will return -1 if the year is less than the minimum reference
year. The problem is that pd.Series[-1] is a valid thing - it returns the 
_last_ value (which is precisely the worst thing to return in this case).

This checks for that and raises.

NOTE: [vivarium does this sort of thing in one location](https://github.com/ihmeuw/vivarium/blob/main/src/vivarium/framework/lookup/interpolation.py#L372) (though it's used on
a series rather than a single value). It handles the case where values fall
below the reference values by grabbing the minimum reference instead
of raising. I'm happy to implement that instead of raising if preferred.

### Testing

We actually had a failing test that I had to update. All tests pass.
I did not create a new test just for this - lmk if you think I should.

### DIFF
--- a/src/vivarium_public_health/population/base_population.py
+++ b/src/vivarium_public_health/population/base_population.py
@@ -166,7 +166,7 @@ class BasePopulation(Component):
         demographic_proportions, year: int
     ) -> pd.DataFrame:
         reference_years = sorted(set(demographic_proportions.year_start))
-        ref_year_index = np.digitize(year, reference_years).item() - 1
+        ref_year_index = _find_bin_start_index(year, reference_years)
         return demographic_proportions[
             demographic_proportions.year_start == reference_years[ref_year_index]
         ]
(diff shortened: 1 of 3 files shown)


src/vivarium_public_health/population/base_population.py
```diff
@@ -257,22 +257,22 @@ def _format_data_inputs(
         if "year_start" not in scaling_factor.index.names:
             return population_structure, scaling_factor
 
-        # Subset to start year of simulation or closest year
```
stevebachmeier, line 260: None of the logic here changed except for calling the new _find_bin_start_index method instead of directly calling np.digitize. Other than that, I just

(1) updated the comment
(2) shuffled the lines around a bit to (I think) make it more clear.

src/vivarium_public_health/population/base_population.py
```diff
@@ -527,3 +527,31 @@ def _assign_demography_with_age_bounds(
...
+        The index of the closest reference value less than or equal to the provided value.
+
+    Raises
+    ------
+        ValueError
+            If the provided value is less than the minimum reference value.
+    """
+    ref_value_index = np.digitize(value, sorted_reference_values).item() - 1
+    if ref_value_index < 0:
+        raise ValueError(
```
stevebachmeier, line 553: See my comment in the description of this PR. Basically, vivarium does a similar thing in one location and if the value falls below the reference values, it chooses the minimum instead of erroring. I'm happy to do that if we think it's better.

stevebachmeier:
> It's probably good to add a test if it's easy to do. Where is the other location we run into this in vivarium?

I linked it in this PR's description: https://github.com/ihmeuw/vivarium/blob/main/src/vivarium/framework/lookup/interpolation.py#L372

hussain-jafari (approved):

PR #640, opened 2026-04-12
fix broken build

rmudambi:
### Description

- *Category*: bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6936

### Changes and notes
Fix broken build 

### Testing
Ran make check

### DIFF
--- a/src/vivarium_public_health/causal_factor/effect.py
+++ b/src/vivarium_public_health/causal_factor/effect.py
@@ -427,7 +427,7 @@ class CausalFactorEffect(Component, ABC):
 
             def generate_relative_risk(index: pd.Index) -> pd.Series:
                 rr = self.relative_risk_table(index)
-                exposure = self.population_view.get_attributes(index, self.exposure_name)
+                exposure = self.population_view.get(index, self.exposure_name)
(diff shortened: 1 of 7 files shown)


src/vivarium_public_health/causal_factor/effect.py
```diff
@@ -436,9 +436,7 @@ def generate_relative_risk(index: pd.Index) -> pd.Series:
 
             def generate_relative_risk(index: pd.Index) -> pd.Series:
                 rr = self.relative_risk_table(index)
-                exposure = self.population_view.get_attributes(
-                    index, self.exposure_name
-                ).reset_index()
+                exposure = self.population_view.get(index, self.exposure_name).reset_index()
```
albrja, line 439: This is weird to me. I see you didn't do this but reset_index seems really scary to do since our index matters and reset_index does not happen above in distribution.py

albrja (approved):

hussain-jafari (approved):

src/vivarium_public_health/causal_factor/effect.py
rmudambi: We're resetting the index so we can properly set it in line 441

PR #648, opened 2026-04-28
population tutorial

hussain-jafari:
population tutorial documentation


docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,840 @@
...
+
+:mod:`vivarium_public_health` provides several components for creating and
+managing simulated populations. This tutorial demonstrates the minimal
+configuration required for each approach.
+
+.. contents::
+   :local:
+   :depth: 2
+
+.. testsetup:: *
```
stevebachmeier, line 13: cool, I didn't know about this directive

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,840 @@
...
+
+   import numpy as np
+   import pandas as pd
+   from vivarium import InteractiveContext
+   from vivarium_public_health.population import *
+   from vivarium_public_health._example_data import *
+   base_plugins = BASE_PLUGINS
```
stevebachmeier, line 20: what is this doing? Where does BASE_PLUGINS come from?

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,840 @@
...
+
+
+Overview
+--------
```
stevebachmeier, line 24: teeny tiny nit, but I like having ==== be the first level header and ----- be the second. It reduces the amount of times you need to resort to weird things like ~~~~ or ^^^^.

Not sure it's worth changing though

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,840 @@
...
+simulation begins:
+
+- :class:`~vivarium_public_health.population.base_population.BasePopulation` — the standard
+  component that samples simulants from demographic data.
+- :class:`~vivarium_public_health.population.base_population.ScaledPopulation` — a variant
+  that rescales the demographic data before sampling.
+
+**Fertility** components add new simulants during the simulation:
+
+- :class:`~vivarium_public_health.population.add_new_birth_cohorts.FertilityDeterministic` — adds a
```
stevebachmeier: god I hate emdashes and endashes. imo we should add to the global instructions to never ever use them and to just use hyphens. I have that in my copilot instructions.

The reason _is_ practical. If a human goes in and adds new documentation, they will presumably use hyphens and so the docs just look inconsistent.

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,840 @@
...
+  fixed number of births per year.
+- :class:`~vivarium_public_health.population.add_new_birth_cohorts.FertilityCrudeBirthRate` — adds
+  births based on a population-level crude birth rate, without accounting for
```
stevebachmeier:
```suggestion
  births based on a population-level crude birth rate without accounting for
```

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,840 @@
...
+  the age or sex composition of the population.
+- :class:`~vivarium_public_health.population.add_new_birth_cohorts.FertilityAgeSpecificRates` —
+  adds births at the individual level based on age-specific fertility rates.
+
+.. note::
+
+   :class:`~vivarium_public_health.population.base_population.BasePopulation` automatically
```
stevebachmeier:
```suggestion
   :class:`~vivarium_public_health.population.base_population.BasePopulation`
```

hussain-jafari:
## PR Review: Population Tutorial (PR #648)

### Summary

This PR adds a new population tutorial (population.rst, 804 lines) and a supporting example data module (_example_data.py, 264 lines). The tutorial covers `BasePopulation`, `ScaledPopulation`, `FertilityDeterministic`, `FertilityCrudeBirthRate`, and `FertilityAgeSpecificRates`, with runnable code blocks, data format documentation, and configuration summary tables. The example data module provides synthetic GBD-shaped DataFrames and an `ExampleArtifactManager` that serves them without requiring an HDF file.

The tutorial is well-structured, progressive, and comprehensive. The example data module is a reasonable approach to making tutorials self-contained. There are several correctness and documentation issues to address before merging.

### Findings

**1. Incorrect location output in "Default configuration" example**
population.rst — The comment shows `location  1`, but the `ExampleArtifactManager` sets `population.location` to `"Kenya"` and `population_structure()` uses `location: "Kenya"`. Vivarium stores the location as a numeric index internally. This inconsistency is never explained and will confuse readers. **Suggestion:** Either show `Kenya` (if that's what actually appears) or add a note explaining the numeric encoding.

**2. Tutorial code blocks use `.. code-block:: python` instead of doctest directives**
Despite `sphinx.ext.doctest` being enabled in conf.py, all tutorial code uses `code-block` which is never executed in CI. Code rot is very likely as components evolve. **Suggestion:** Either convert to `.. testsetup::` / `.. testcode::` / `.. testoutput::`, or add a separate integration test that exercises the key tutorial scenarios programmatically. *(HIGH priority)*

**3. RST substitution `|times|` used in Python docstring**
_example_data.py — The `live_births_by_sex()` docstring contains `|times|` which is an RST substitution defined in the `.rst` file but not available in Python docstrings. It renders literally as `|times|` in `help()` and IDEs. **Suggestion:** Use the Unicode character `×` directly or the word "by".

**4. `Columns:` lines placed outside any NumPy doc section**
_example_data.py and _example_data.py — In `live_births_by_sex()` and `age_specific_fertility_rate()`, the `Columns:` documentation appears after the `Parameters` section but outside any recognized NumPy doc section. Sphinx's napoleon extension won't parse these correctly. **Suggestion:** Move the `Columns:` info into a `Returns` section or into the extended summary above `Parameters`.

**5. Missing CHANGELOG entry**
CHANGELOG.rst has no entry for this PR. Following recent entries (`Add treatment API docs`, `Add risk concept docs`, etc.), an entry like "Add population tutorial and example data module" is expected.

**6. `is_aged_out` column introduced without explanation**
population.rst — The "Untracking age" example uses `is_aged_out` in `get_population()` without explaining where this column comes from or when it's available. **Suggestion:** Add a brief note (e.g., "The `is_aged_out` column is populated by the `AgeOutSimulants` sub-component").

**7. `FertilityCrudeBirthRate` example writes data already provided by `ExampleArtifactManager`**
population.rst — The `_ARTIFACT_DATA` dict in `_example_data.py` already includes `"covariate.live_births_by_sex.estimate": live_births_by_sex`. The explicit `sim._data.write(...)` is redundant and could confuse readers into thinking the write is required. **Suggestion:** Either (a) remove the write and simplify to `setup=True`, or (b) add a note: "The example artifact already provides this key; the explicit write below demonstrates the pattern you would use with a real artifact."

**8. Module docstring says "round numbers" but age bins aren't**
_example_data.py — The module docstring says "intentionally simple (uniform rates, round numbers)" but the neonatal boundaries (0.01917808 = 7/365, 0.07671233 = 28/365) are precise GBD conventions, not round numbers. **Suggestion:** Amend to distinguish: "uniform rates and round population counts; age bins follow standard GBD definitions."

### Overall

This is a well-crafted, comprehensive tutorial that fills a real gap in the documentation. The example data module is thoughtfully designed to ship with the package. The **highest-priority items** to address before merging are:

1. **Make tutorial code blocks testable in CI** (finding #2) — this is the only guard against code rot since `_example_data` is tutorial-only.
2. **Add a CHANGELOG entry** (finding #5).
3. **Fix the `|times|` RST substitution** in the Python docstring (finding #3) — renders incorrectly in all non-RST contexts.
4. **Clarify or fix** the `FertilityCrudeBirthRate` redundant write pattern (finding #7) — as written, it teaches an unnecessary step.
5. **Address the location output discrepancy** (finding #1) — `1` vs. `"Kenya"`.

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,840 @@
...
+   includes three sub-components:
+   :class:`~vivarium_public_health.population.mortality.Mortality`,
+   ``AgeOutSimulants``, and ``Disability``. You do not need to add these
```
stevebachmeier:
```suggestion
   ``AgeOutSimulants``, and ``Disability``. Do not explicitly add these to the model specification.
```

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,845 @@
...
+   from vivarium_public_health._example_data import *
+   base_plugins = BASE_PLUGINS
+
+
+Overview
+--------
+
+There are two categories of population components:
+
+**Initial population** components create the starting set of simulants when the
```
albrja, line 28: Base population?

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,840 @@
...
+- :class:`~vivarium_public_health.population.add_new_birth_cohorts.FertilityAgeSpecificRates` —
+  adds births at the individual level based on age-specific fertility rates.
+
+.. note::
+
+   :class:`~vivarium_public_health.population.base_population.BasePopulation` automatically
+   includes three sub-components:
+   :class:`~vivarium_public_health.population.mortality.Mortality`,
+   ``AgeOutSimulants``, and ``Disability``. You do not need to add these
+   separately.
```
stevebachmeier:
```suggestion
```

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,840 @@
...
+
+
+Common Setup
+------------
+
+In a production simulation, data is supplied through a **data artifact** —
+an HDF file that you build with all the input data your model needs.  To
```
stevebachmeier: nit: replace all double-spaces-between-sentences with single spaces

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,840 @@
...
+   includes three sub-components:
+   :class:`~vivarium_public_health.population.mortality.Mortality`,
+   ``AgeOutSimulants``, and ``Disability``. You do not need to add these
+   separately.
+
+
+Common Setup
+------------
+
+In a production simulation, data is supplied through a **data artifact** —
```
stevebachmeier:
```suggestion
In a vivarium simulation, data can be supplied through a **data artifact** —
```

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,845 @@
...
+^^^^^^^^^
+
+The table below lists every data key used by the population components.
+Keys marked **artifact-required** must be present in the artifact — the
+component loads them directly and they cannot be replaced via configuration.
+Keys marked **configurable** can be overridden in the ``data_sources``
+section of the configuration (see `Data sources`_); the artifact key shown
+is simply the default.
+
+.. list-table::
```
stevebachmeier, line 102: This kind of stuff scares me. It's easily generated by ai, but will quickly go out of date. 🤷

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,845 @@
...
+     - Yes |mdash| ``fertility_age_specific_rates.data_sources.age_specific_fertility_rate``
+
+
+.. |mdash| unicode:: U+2014
+
+
+Data sources
+^^^^^^^^^^^^
+
+Some components support a ``data_sources`` configuration pattern that lets
```
stevebachmeier, line 151: don't _all_ components?

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,845 @@
...
+.. code-block:: yaml
+
+   # Default configuration (loaded from the artifact):
+   mortality:
+     data_sources:
+       all_cause_mortality_rate: "cause.all_causes.cause_specific_mortality_rate"
+       life_expectancy: "population.theoretical_minimum_risk_life_expectancy"
+       unmodeled_cause_specific_mortality_rate: <internal method>
+
+Note: The ``unmodeled_cause_specific_mortality_rate`` default is shown as
```
stevebachmeier: you can put notes via the `.. note::` directive

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,845 @@
...
+   from vivarium_public_health._example_data import (
+       population_structure,
+       age_bins,
+       theoretical_minimum_risk_life_expectancy,
+   )
+
+   # population.structure — population counts per demographic cell.
+   pop_structure = population_structure()
+   print(pop_structure.query("year_start == 1990").head(6).to_string(index=False))
+   #  age_start  age_end    sex  year_start  year_end location      value
```
stevebachmeier: instead of adding a commented block of what the output should look like, add a `.. testoutput::` directive to actually test is. (and it shows up in the rendered docs as well)

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,845 @@
...
+   tmrle = theoretical_minimum_risk_life_expectancy()
+   print(tmrle.head(5).to_string(index=False))
+   #  age_start  age_end  value
+   #        0.0      1.0   98.0
+   #        1.0      2.0   98.0
+   #        2.0      3.0   98.0
+   #        3.0      4.0   98.0
+   #        4.0      5.0   98.0
+
+.. testoutput::
```
stevebachmeier, line 270: this doesn't seem correct.

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,845 @@
...
+       components=[BasePopulation()],
+       configuration=config,
+       plugin_configuration=base_plugins,
+   )
+
+   pop = sim.get_population(["age"])
+   print(f"Age range: {pop['age'].min():.2f} – {pop['age'].max():.2f}")
+   # Age range: 0.00 – 4.99  (approximate; all simulants are between 0 and 5)
+
+.. testoutput::
```
stevebachmeier, line 355: none of these should be here I don't think. what is it actually testing?

To clarify, I think what you want is:

```
.. testcode::

    # set everything up
    print(XXX)

.. testoutput::

   <stdout from XXX>

.. testcode::

    # set the next bit up
    print(YYY)

.. testoutput::

    <stdout from YYY>

.
.
.
```

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,845 @@
...
+   # Female    10000
+
+.. testoutput::
+   :hide:
+   :options: +ELLIPSIS, +NORMALIZE_WHITESPACE
+
+   ...
+
+
+Untracking age (aging out)
```
stevebachmeier:
```suggestion
Aging out of a simulation
```

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,845 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Setting ``untracking_age`` causes simulants to be removed from the tracked
```
stevebachmeier, line 436: You should link to the higher-level vivarium docs that talk about how untracking works (if there are any...I think there are?)

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,845 @@
...
+   # Aged out: 10000
+
+.. testoutput::
+   :hide:
+   :options: +ELLIPSIS, +NORMALIZE_WHITESPACE
+
+   ...
+
+
+Configuration summary for BasePopulation
```
stevebachmeier, line 487: A lot of the docs here seem like they should be more part of the concept docs rather than tutorials. I'm not 100% sure though? But like...isn't all of this in the concept docs as well? We def don't want repeated documentation and can discuss where the line is between concept and tutorial as a team.

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,845 @@
...
+
+   ...
+
+Because this component supports the ``data_sources`` configuration, the
+tutorial example below supplies a constant rate directly instead of loading
+from the artifact:
+
+.. testcode::
+
+   from vivarium import InteractiveContext
```
stevebachmeier, line 791: note: you don't really need to keep importing things over and over - everything persists from one test to the next. It kind of depends on what you want the user to see. If the intent is one long continuous "session", then don't keep importing. If you want each section to show from scratch, then do.

src/vivarium_public_health/_example_data.py
```diff
@@ -0,0 +1,306 @@
+"""
```
stevebachmeier, line 1: This shouldn't be in src if we can help it. Any way to leverage it from within the docs/ dir?

stevebachmeier (changes requested):
Looks good! My main issues are
1. I don't think example data belongs in src/
2. You're not really testing the output the way you think you are (I don't think, unless I don't understand what the elipses stuff is doing).

albrja (approved):

docs/source/tutorials/population.rst
hussain-jafari: Added comments.

docs/source/tutorials/population.rst
hussain-jafari: No, some of them need an artifact because they call builder.data.load(key) explicitly.

docs/source/tutorials/population.rst
hussain-jafari: It's not real data, but the structure of the data is what matters.

docs/source/tutorials/population.rst
hussain-jafari: I'm imagining a user might only be looking at one component at a time.

docs/source/tutorials/population.rst
hussain-jafari: The concept docs don't mention most key names or the expected data structure and when they do it's not in easily referenced/structured way. There are occasionally 1 to 2 sentence overlaps when describing components.

src/vivarium_public_health/_example_data.py
hussain-jafari: We would then have to add the docs dir to sys.path to import the ExampleArtifactManager.

docs/source/tutorials/population.rst
```diff
@@ -212,52 +226,62 @@ the same column layout but with real GBD values.
...
-   #   0.000000 0.019178 Female        1990      1991    Kenya   1.917808
-   #   0.019178 0.076712   Male        1990      1991    Kenya   5.753425
-   #   0.019178 0.076712 Female        1990      1991    Kenya   5.753425
-   #   0.076712 1.000000   Male        1990      1991    Kenya  92.328767
-   #   0.076712 1.000000 Female        1990      1991    Kenya  92.328767
-
-   # population.age_bins — defines the age groups used by the demographic data.
+
+.. testoutput::
+   :options: +NORMALIZE_WHITESPACE
```
stevebachmeier, line 234: omg if this does what I think/hope it does then that's amazing

docs/source/tutorials/population.rst
stevebachmeier: I'm referring to the `.. testoutput::`. Looks like you fixed though

docs/source/tutorials/population.rst
stevebachmeier: I mean, like..._where_ does BASE_PLUGINS come from? Is it some sphinx doc auto-imported thing or somethign like that? I don't see it imported from anywhere or defined anywhere.

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,864 @@
...
+   assert len(pop) == 10_000
+   assert pop["age"].min() >= 0
+   assert pop["age"].max() <= 125
+   assert set(pop["sex"].unique()) == {"Male", "Female"}
+   print(f"Population: {len(pop)}, ages {pop['age'].min():.1f}–{pop['age'].max():.1f}")
+
+.. testoutput::
+   :options: +ELLIPSIS, +NORMALIZE_WHITESPACE
+
+   Population: 10000, ages ...
```
stevebachmeier: testing hidden output (...) isn't useful. Change the print statement so it actually prints both, e.g. something like

```
print(len(pop))
print(pop['age'].min())
print(pop['age'].max())
```

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,864 @@
...
+   # After running for 600 days (~1.6 years), everyone has aged past 5
+   import pandas as pd
+   sim.run_for(duration=pd.Timedelta(days=600))
+   pop = sim.get_population(["is_aged_out", "exit_time"], include_untracked=True)
+   print(f"Aged out: {pop['is_aged_out'].sum()}")
+
+.. testoutput::
+   :options: +ELLIPSIS, +NORMALIZE_WHITESPACE
+
+   ...
```
stevebachmeier: Maybe I'm not understanding these ellipses. Why are they here? Why isn't the output just `Aged out: 10000`?

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,864 @@
...
+   sim.run_for(duration=pd.Timedelta(days=100))
+   pop = sim.get_population(["age"])
+   # Population grew from 1000 by ~500 * (100/365) ≈ 137 new simulants.
+   assert len(pop) > 1_000
+   print(f"Population grew: {len(pop) > 1_000}")
+
+.. testoutput::
+   :options: +ELLIPSIS, +NORMALIZE_WHITESPACE
+
+   ...
```
stevebachmeier: ellipsis

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,864 @@
...
+   import pandas as pd
+   sim.run_for(duration=pd.Timedelta(days=100))
+   pop = sim.get_population(["age"])
+   assert len(pop) > 10_000
+   print(f"Population grew: {len(pop) > 10_000}")
+
+.. testoutput::
+   :options: +ELLIPSIS, +NORMALIZE_WHITESPACE
+
+   ...
```
stevebachmeier: ellipsis

docs/source/tutorials/population.rst
```diff
@@ -0,0 +1,864 @@
...
+
+   # Newborns have a parent_id pointing to their mother
+   newborns = pop[pop["parent_id"] >= 0]
+   assert len(newborns) > 0
+   print(f"Births occurred: {len(newborns) > 0}")
+
+.. testoutput::
+   :options: +ELLIPSIS, +NORMALIZE_WHITESPACE
+
+   ...
```
stevebachmeier: ellipsis

stevebachmeier (approved):

docs/source/tutorials/population.rst
hussain-jafari: The ages are non-deterministic so we just print length. We test that they're in the expected range earlier.

docs/source/tutorials/population.rst
hussain-jafari: There's some extra output produced by sim.run_for() that we don't want to look at.

PR #649, opened 2026-04-30
Hjafari/docs/mic 6886 disease tutorial

hussain-jafari:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-XYZ


hussain-jafari:
## PR Review: Add disease tutorial

### Summary

This PR adds a comprehensive 1345-line RST tutorial (disease.rst) demonstrating disease model components in `vivarium_public_health`, along with supporting example data functions in _example_data.py. The tutorial covers manual model construction, pre-built models (SI, SIS, SIR, fixed-duration, neonatal), and advanced topics (dwell time, excess mortality, proportion/transient transitions, sequelae, config overrides, event tracking). The _example_data.py changes add disease data builder functions, tutorial-specific artifact overrides, and a fallback lookup mechanism for arbitrary cause keys.

### Design

1. **Over-broad fallback matching in `_ExampleArtifact.load()`** — _example_data.py

   The fallback splits on `"."` and checks `len(parts) >= 2`, matching `parts[-1]` against `_CAUSE_DEFAULTS`. This means *any* dotted key whose last segment is a known measure name would match — e.g., `"covariate.something.prevalence"` or `"risk.smoking.incidence_rate"`. The comment says "cause.{name}.{measure} keys" but the code doesn't enforce it.

   **Suggestion:** Tighten to require the `cause.` prefix and exactly 3 parts:
   ```python
   parts = entity_key.split(".")
   if len(parts) == 3 and parts[0] == "cause":
       measure = parts[2]
       if measure in _CAUSE_DEFAULTS:
           value = _CAUSE_DEFAULTS[measure]
           return value() if callable(value) else value
   ```

2. **Separate wrapper functions are the right choice for this context** — _example_data.py

   The five one-line wrappers (`disease_prevalence`, `disease_incidence_rate`, etc.) are appropriate because they provide distinct public names with domain-appropriate defaults, are imported individually in the tutorial, and are referenced individually in `_CAUSE_DEFAULTS`. No change needed.

### Maintainability

1. **Misleading parameter name `rate` for `disease_prevalence()`** — _example_data.py

   Prevalence is a proportion (bounded 0–1), not a rate (which can exceed 1). The parameter is named `rate`, which contradicts the docstring's "The constant prevalence applied to…". This also affects the call site `disease_prevalence(rate=0.05)` for birth_prevalence at _example_data.py.

   **Suggestion:** Rename to `proportion` or `value`:
   ```python
   def disease_prevalence(proportion: float = 0.0) -> pd.DataFrame:
   ```

2. **Module docstring only references population tutorial** — _example_data.py

   The docstring says "See the :doc:`/tutorials/population` tutorial for usage" but the module now also serves the disease tutorial.

   **Suggestion:** Add `:doc:`/tutorials/disease`` to the reference list.

3. **Magic rate values without explanatory comments** — _example_data.py

   The rates 0.5, 5.0, 0.05, 1.0 are tuned to produce visible transitions in tutorial steps, but there's no comment explaining why these values were chosen.

   **Suggestion:** Add brief inline comments (e.g., `# High enough to guarantee transitions in 5-10 steps`).

### DRY

1. **Duplicated age×sex×year grid construction** — _example_data.py and _example_data.py

   `_build_cause_table()` and `population_structure()` share the same grid-building pattern (product of bins/sexes/years, unpacking tuples). A shared `_age_sex_year_grid()` helper would eliminate duplication. This is medium severity — a change to the grid structure would require updating both places.

   **Suggestion:** Extract a shared grid builder:
   ```python
   def _age_sex_year_grid() -> pd.DataFrame:
       rows = list(product(_AGE_BINS, ("Female", "Male"), _YEAR_BINS))
       mins, maxes = zip(*[r[0] for r in rows])
       sex_col = [r[1] for r in rows]
       y_starts, y_ends = zip(*[r[2] for r in rows])
       return pd.DataFrame({
           "age_start": mins, "age_end": maxes,
           "sex": sex_col, "year_start": y_starts, "year_end": y_ends,
       })
   ```

2. **Tutorial config repetition is acceptable** — The ~15 repeated config blocks in disease.rst are fine since `testcode` blocks should be independently runnable and `make_base_config()` already extracts the shared core.

### Tests

1. **No unit tests for `_ExampleArtifact.load()` fallback logic** — _example_data.py

   The fallback parsing has branching logic (split, length check, measure lookup) that's only tested indirectly through the full doctest pipeline. A regression in the parsing wouldn't be caught until the slow doc build runs.

   **Suggestion:** Add a small unit test for the fallback mechanism covering: novel cause resolution, unknown measure raises KeyError, and write-override precedence.

2. **SIR fixed-duration test doesn't assert the infected state exists** — disease.rst

   The test only verifies `susceptible` and `recovered` states are present. If simulants skipped directly to recovered, the test would still pass.

   **Suggestion:** Add an assertion for `"test_cause" in states` or verify at an intermediate step.

3. **Weak `n_infected > 0` assertions** — disease.rst

   With rate=0.5 and 10,000 simulants over 5 steps, `> 0` would pass even with a 99% regression in incidence. Acceptable for a doctest (avoiding flakiness), but a tighter bound like `> 100` would still be safe and catch catastrophic regressions.

### Documentation

1. **Hidden imports in `testsetup` not shown to users** — disease.rst

   The `testsetup` block does `from vivarium_public_health.disease import *` but this is invisible to users. The "Common Setup" section only shows `BASE_PLUGINS` and `make_base_config`. Users copying examples into standalone scripts will get `NameError` for `SusceptibleState`, `DiseaseState`, etc.

   **Suggestion:** Add an explicit import block in the visible "Common Setup" section:
   ```rst
   .. testcode::

      from vivarium_public_health.disease import *
   ```

2. **SIR section omits shared artifact keys** — disease.rst

   The "Artifact keys used" list only mentions `incidence_rate` and `remission_rate`, but SIR also requires prevalence, disability_weight, excess_mortality_rate, and CSMR. The Configuration Summary table at the bottom is correct, creating an inconsistency.

   **Suggestion:** Add the full key list (or say "In addition to the keys listed for SI:").

3. **Section title style inconsistency** — disease.rst

   "Using Pre-Built Models" uses a gerund, while all other sections use noun phrases ("DiseaseModel", "Neonatal Models", "Advanced Topics").

   **Suggestion:** Rename to "Pre-Built Models" for consistency.

### Functionality

1. **`_CAUSE_DEFAULTS` stores callables with zero-arg signatures, but `disease_prevalence`/`disease_incidence_rate` etc. have *optional* parameters** — _example_data.py

   The fallback always calls `value()` with no arguments, so it always gets the *default* values (prevalence=0.0, incidence=0.001, etc.). This is correct and intentional — the defaults serve as "generic" sensible values while `_ARTIFACT_DATA` stores tutorial-specific overrides with higher rates. No bug here, but worth confirming this was deliberate since a `cause.new_disease.incidence_rate` fallback will always return rate=0.001 regardless of context.

2. **`disease_prevalence(rate=0.05)` in `_ARTIFACT_DATA` for `neonatal_cause.birth_prevalence` is semantically correct but naming is confusing** — _example_data.py

   Passing `rate=0.05` for a prevalence function that represents a proportion is technically safe (0.05 < 1), but if someone mistakenly passes a large value like `rate=5.0` (thinking it's a rate), the resulting "prevalence" DataFrame would contain invalid fractions. No validation exists.

3. **Pre-existing issue in `_ARTIFACT_DATA`: sex ordering differs between `population_structure` and `_build_cause_table`** — _example_data.py vs _example_data.py

   `population_structure()` uses `("Male", "Female")` while `_build_cause_table()` uses `("Female", "Male")`. This doesn't affect correctness (order within the DataFrame doesn't matter for lookup), but could cause confusion in doctest output if someone adds assertions about row ordering.

### Minor Nits

1. **Em-dash (—) in existing code** — _example_data.py: The comment `# Mortality data — zero by default` uses an em-dash. Per team style, this should be a hyphen. (Pre-existing, not introduced by this PR.)

2. **`_CAUSE_DEFAULTS` type annotation** — _example_data.py: `dict[str, object]` doesn't communicate that values must be callables. Consider `dict[str, Callable[[], Any]]` for clarity.

### Overall

This is a well-structured, comprehensive tutorial that covers the disease modeling API end-to-end. The code additions are clean and appropriately scoped for tutorial-only use. The two most actionable items are:

1. **Tighten the fallback key matching** (Design #1) — a one-line fix that prevents silent incorrect data serving for non-cause keys.
2. **Show the disease import in the visible tutorial text** (Documentation #1) — users copying examples will be blocked without it.

The parameter naming (`rate` vs `proportion`) is a recurring theme flagged by multiple review perspectives and would be worth addressing for clarity, though it doesn't cause bugs in practice since all tutorial values are < 1.

docs/nitpick-exceptions
```diff
@@ -41,3 +41,22 @@ py:attr lookup_tables
 
 # typing module - needed for Python 3.10 compatibility
 py:data typing.Union
+
+# disease tutorial - references use package-level paths for readability
```
stevebachmeier: I actually wouldn't do this since that means the user can't do what the tutorial does

src/vivarium_public_health/_example_data.py
```diff
@@ -64,10 +67,27 @@
 
 
 ##############################################
-# Data builders — one per artifact key shape #
+# Data builders - one per artifact key shape #
 ##############################################
 
 
+def _age_sex_year_grid() -> pd.DataFrame:
```
stevebachmeier: i don't like this name. make it a `get_` method, e.g. `_get_age_sex_year_combos` or something

src/vivarium_public_health/_example_data.py
```diff
@@ -198,6 +202,140 @@ def age_specific_fertility_rate(rate: float = 0.05) -> pd.DataFrame:
     )
 
 
+########################
+# Disease data helpers #
+########################
+
+
+def disease_prevalence(rate: float = 0.0) -> pd.DataFrame:
```
stevebachmeier: This is super WET. Would it make sense to collapse into a single function?

docs/source/tutorials/disease.rst
```diff
@@ -0,0 +1,1323 @@
+==================================
+Disease Models and State Machines
+==================================
+
+:mod:`vivarium_public_health` provides a flexible framework for modelling
+diseases as state machines. This tutorial demonstrates how to build disease
```
stevebachmeier, line 6: link to any state machine docs we may have

docs/source/tutorials/disease.rst
```diff
@@ -0,0 +1,1323 @@
...
+models from states and transitions, and how to use the pre-built models for
```
stevebachmeier, line 7: link to state and transition docs

docs/source/tutorials/disease.rst
```diff
@@ -0,0 +1,1323 @@
...
+   from vivarium_public_health.population import BasePopulation
+   from vivarium_public_health._example_data import *
+   base_plugins = BASE_PLUGINS
+
+
+Overview
+--------
+
+A disease model in ``vivarium_public_health`` is a state machine. Each
+simulant occupies exactly one disease state at any time and moves between
```
stevebachmeier: This is kinda ambiguous. A simulant can occupy multiple disease states _from different diseases_ at the same time

docs/source/tutorials/disease.rst
```diff
@@ -0,0 +1,1323 @@
...
+states according to transition rules.
+
+The building blocks are:
```
stevebachmeier: This all feels very duplicative of the concept docs. But I thought that last time as well. Just make sure this is DRY and reference those docs instead of re-writing here.

docs/source/tutorials/disease.rst
```diff
@@ -0,0 +1,1323 @@
...
+
+.. testcode::
+
+   # cause.{cause}.incidence_rate - rate of new infections per person-year.
+   # Same column layout as prevalence.
+   incidence = disease_incidence_rate(0.001)
+   print(incidence.query("year_start == 1990").head(2).to_string(index=False))
+
+.. testoutput::
+   :options: +NORMALIZE_WHITESPACE
```
stevebachmeier, line 183: does the NORMALIZE_WHITESPACE do what I think/hope it does - does it mean that we don't need to get the spacing _exactly_ correct? That was always super annoying when updating these testoutputs.

docs/source/tutorials/disease.rst
```diff
@@ -0,0 +1,1323 @@
...
+
+
+Building a model from scratch
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The most explicit way to create a disease model is to instantiate states,
+wire up transitions, and wrap them in a
+:class:`~vivarium_public_health.disease.DiseaseModel`.
+
+The following example builds an SIS (Susceptible |harr| Infected |harr|
```
stevebachmeier, line 313: what is `|harr|`?

docs/source/tutorials/disease.rst
```diff
@@ -0,0 +1,1323 @@
...
+
+- ``cause.{cause}.incidence_rate`` - for the susceptible |rarr| infected transition
+- ``cause.{cause}.prevalence`` - for initialization
+- ``cause.{cause}.disability_weight`` - for YLD calculation
+- ``cause.{cause}.excess_mortality_rate`` - for mortality calculation
+- ``cause.{cause}.cause_specific_mortality_rate`` - for CSMR
+
+.. testcode::
+
+   config = make_base_config()
```
stevebachmeier, line 463: Wait, where do all of the SI details then come from? Like...the comment above says that it uses various keys from the artifact, but where did that artifact get created? Is that from `make_base_config()`? If so, I think we need to somehow direct the users to that. B/c right now it kinda seems like these models just magically require no input data.

docs/source/tutorials/disease.rst
```diff
@@ -0,0 +1,1323 @@
...
+
+Neonatal disease models assign a condition at birth based on birth
+prevalence. They are designed for conditions that are present from the
+start of life.
+
+**Artifact keys used:**
+
+- ``cause.{cause}.birth_prevalence`` - for assigning condition at birth
+
+NeonatalSWC without incidence
```
stevebachmeier, line 706: What is SWC? Make sure to add the answer wherever appropriate to the docs

docs/source/tutorials/disease.rst
```diff
@@ -0,0 +1,1323 @@
...
+
+
+Advanced Topics
+----------------
+
+
+Dwell time
+^^^^^^^^^^
+
+A **dwell time** forces simulants to remain in a state for a minimum
```
stevebachmeier, line 819: Should all/most/some of these terms be getting added to a public health glossary?

docs/source/tutorials/disease.rst
```diff
@@ -0,0 +1,1323 @@
...
+
+.. testoutput::
+   :options: +ELLIPSIS
+
+   ...
+   No simulants in router: True
+   Both outcomes populated: True
+
+
+Multiple disease states (sequelae)
```
stevebachmeier, line 1067: This is in direct conflict with that very first sentence saying a simulant can only be in one state at a time.

stevebachmeier (approved):
Approved, though I do think there's some improvements to be made

albrja (approved):

docs/source/tutorials/disease.rst
hussain-jafari: We don't have any :/

docs/source/tutorials/disease.rst
hussain-jafari: I agree this time this is all duplicative and redundant. Replaced with a reference to the concept docs.

docs/source/tutorials/disease.rst
hussain-jafari: Yup :)

docs/source/tutorials/disease.rst
hussain-jafari: Double sided arrow

docs/source/tutorials/disease.rst
hussain-jafari: base_plugins uses the our custom artifact manager to custom load data. I've updated the documentation to accurately reflect this.

docs/source/tutorials/disease.rst
hussain-jafari: That statement was for a given disease model, and has been updated to clarify that.

docs/source/tutorials/disease.rst
hussain-jafari: They have :)

PR #650, opened 2026-05-04
risk tutorial

hussain-jafari:
risk tutorial


hussain-jafari:
I now have enough context for a complete review. Here it is:

---

## PR Review: risk tutorial (#650)

### Summary

This PR adds a comprehensive risk factor tutorial (risk.rst, 771 lines) and supporting data helpers in _example_data.py (~117 added lines). The tutorial covers `Risk`, `RiskEffect`, and `NonLogLinearRiskEffect` components with runnable doctest examples. The `_example_data` additions follow the existing pattern from the disease/population tutorials - adding helper functions, `_ARTIFACT_DATA` entries, and a `_RISK_DEFAULTS` fallback dict.

### Design

1. **`_RISK_DEFAULTS` fallback produces entity-unaware data** — _example_data.py

   `_RISK_DEFAULTS["relative_risk"]` calls `risk_relative_risk_dichotomous()` which always returns `affected_entity="test_cause"`. If a tutorial uses `risk_factor.smoking.relative_risk` via the fallback (without scalar override), it gets data pointing at `"test_cause"` rather than the actual target. This is a latent correctness trap - consider adding a comment noting the limitation, or making the callables entity-aware (accepting `parts[1]`).

2. **`ContinuousExposureRisk` subclass teaches an atypical pattern** — risk.rst

   The tutorial defines a custom `Risk` subclass that completely bypasses parent machinery (propensity, distribution lookup, framework randomness). While pragmatic for demonstrating `NonLogLinearRiskEffect`, readers may adopt this as a production pattern. Add a `.. note::` stating this is a minimal demo shortcut, and that production continuous risks should use `Risk` with a continuous `distribution_type`.

3. **`sim._data.write()` accesses private API without caveat** — risk.rst

   The `setup=False` workflow inherently requires writing to the artifact before setup. Consider adding a note that `_data` is internal and this pattern is specific to interactive/tutorial contexts.

### Maintainability

4. **Fragile key-parsing in `_ExampleArtifact.load()`** — _example_data.py

   The fallback checks `parts[0] == "risk_factor"` but doesn't handle `"alternative_risk_factor"` (a valid entity type in the framework). A future tutorial author would get an opaque `KeyError`. Add a comment noting the supported prefixes.

5. **Hardcoded column name couples tutorial to internal naming** — risk.rst

   `exposure_col = "test_risk_exposure_for_non_loglinear_riskeffect"` mirrors an internal naming convention. If it changes, this tutorial breaks silently. Add a brief comment referencing where this name comes from.

6. **`_RISK_DEFAULTS["tmred"]` schema undocumented** — _example_data.py

   The dict `{"distribution": "uniform", "min": 0, "max": 0, "inverted": False}` has no inline explanation. Add a comment like `# min/max define the no-risk exposure range; inverted=True means lower exposure is riskier`.

### DRY

7. **`risk_exposure_dichotomous` and `risk_relative_risk_dichotomous` share a dichotomous-split pattern** — _example_data.py

   Both copy the grid, add "exposed"/"unexposed" parameter rows, and concat. Consider extracting a `_build_dichotomous_table(exposed_value, unexposed_value, extra_columns=None)` helper, analogous to the existing `_build_cause_table`.

8. **RR data construction duplicated in tutorial** — risk.rst and risk.rst

   The 1000-threshold `rr_data` DataFrame is constructed identically in both "Building relative risk data" and "Running with continuous exposure". Since these are separate testcode blocks this is expected, but consider providing a `risk_relative_risk_continuous()` helper in _example_data.py to reduce to a one-liner and centralize the data shape.

### Tests

9. **NonLogLinearRiskEffect verification too weak** — risk.rst

   The test only asserts `high_rate > low_rate` and `ratio > 1.0`. A broken interpolation returning a uniform small RR could still pass. Consider tightening to something like `1.3 < ratio < 3.0`.

10. **Multiple risk effects - tight stochastic bounds for small subgroups** — risk.rst

    The `pollution_only` group (~2,100 simulants) and `neither_exposed` group (~2,100) have the tightest tolerance (`1.5 < pollution_ratio < 2.5`). Consider widening to `1.3 < pollution_ratio < 2.7` or increasing `population_size` to reduce CI flakiness risk.

11. **PAF test uses single step** — risk.rst

    `run_paf_sim` runs only 1 step, producing ~420 infections per run. The ratio of two small rates has high variance. Running 2-3 steps would improve reliability.

### Documentation

12. **Data keys table incomplete for RR/PAF** — risk.rst

    The "Index columns" for `relative_risk` lists "age, sex, year, parameter" but the actual DataFrame also includes `affected_entity` and `affected_measure` columns (visible in the testoutput below). Add these or note that "age" and "year" are shorthand for start/end pairs.

13. **Module docstring doesn't reference risk tutorial** — _example_data.py

    The docstring references population and disease tutorials but not the new risk tutorial. Update to include `:doc:`/tutorials/risk``.

14. **Inconsistent `{target}` template in YAML examples** — risk.rst vs risk.rst

    The default config uses `risk_effect.{name}_on_{target}:` but the override section uses `risk_effect.{name}_on_{target_entity}.{target_name}.{target_measure}:`. Add a one-time note that `{target}` expands to the dotted form.

### Functionality

15. **Dead import `from vivarium import Component`** — risk.rst

    `Component` is imported but never used — the class inherits from `Risk`. Remove this import.

16. **Magic tolerance ranges lack derivation** — risk.rst

    The first occurrence of a stochastic tolerance (`3.0 < ratio < 7.0` for expected RR=5) should have a comment explaining the reasoning, e.g., `# ±40% tolerance accounts for stochastic variation with N=10,000 over 3 steps`.

### Minor Nits

17. `testsetup` imports `NonLogLinearRiskEffect` from `risks`, but the NonLogLinearRiskEffect testcode block redundantly re-imports from `risks.effect` (line 642). Harmless but confusing to readers.

18. No test for the identity case (RR=1 should produce equal rates between exposed/unexposed). Low priority but would complete the edge-case coverage.

### Overall

This is a well-structured, pedagogically sound tutorial that follows the patterns established in the population and disease tutorials. The primary concerns are: (1) the NonLogLinearRiskEffect example's weak assertion that could pass with broken interpolation, (2) undocumented coupling to internal naming conventions, and (3) some tight stochastic bounds that could cause CI flakiness. The `_example_data` additions are appropriate and correctly scoped to tutorial use only. Most findings are low-to-medium severity polish items.

docs/source/tutorials/risk.rst
```diff
@@ -0,0 +1,759 @@
...
+
+    age_start   age_end    sex  year_start  year_end parameter  value
+     0.000000  0.019178   Male        1990      1991   exposed    0.6
+     0.000000  0.019178 Female        1990      1991   exposed    0.6
+     0.019178  0.076712   Male        1990      1991   exposed    0.6
+     0.019178  0.076712 Female        1990      1991   exposed    0.6
+
+.. testcode::
+
+   # risk_factor.{name}.relative_risk - RR per exposure category and target.
```
albrja: I think you should mention target above when you introduce risk effects.

docs/source/tutorials/risk.rst
```diff
@@ -0,0 +1,759 @@
...
+RiskEffect
+----------
+
+A :class:`~vivarium_public_health.risks.effect.RiskEffect` modifies disease
+dynamics based on exposure. In the standard pattern, exposed simulants have
+a higher incidence rate (multiplied by the relative risk) than unexposed
+simulants.
+
+
+Observing the effect
```
albrja: I think it would be better to split exposure and effect into different modules. You could even say make sure to go through exposure first.

docs/source/tutorials/disease.rst
albrja, line 219: Nit: This is'nt limited to disease components.

albrja, line 438: IMO it would be more helpful to new users to have the pre-built models section before building a model from scratch. You can show them these items exist and you can just use them, but if you wanted to do something more complicated, you can build a model from scratch. Above you build an SIS model and if they have the context for what that is before they build it then you are presenting less information at once to the reader.

docs/source/tutorials/risk.rst
```diff
@@ -0,0 +1,759 @@
+===============================
+Risk Factors and Risk Effects
```
albrja: I think this needs to be updated?

docs/source/tutorials/risk.rst
```diff
@@ -0,0 +1,759 @@
...
+   proportion = n_exposed / len(pop)
+   # Approximately 40% should be exposed.
+   print(f"Proportion exposed near 0.4: {0.35 < proportion < 0.45}")
+
+.. testoutput::
+
+   Proportion exposed near 0.4: True
+
+
+RiskEffect
```
albrja: Is this still suppose to be in this file?

docs/source/tutorials/risk.rst
```diff
@@ -0,0 +1,759 @@
...
+----------
+
+A :class:`~vivarium_public_health.risks.effect.RiskEffect` modifies disease
```
albrja: Yea this is just repeated in the effect module now.

albrja (changes requested):

docs/source/tutorials/risk.rst
```diff
@@ -0,0 +1,282 @@
...
+
+
+Overview
+--------
+
+A risk model in ``vivarium_public_health`` has two primary components:
+
+**Exposure** - the risk factor and how simulants are exposed to it:
+
+- :class:`~vivarium_public_health.risks.base_risk.Risk` - assigns each
```
stevebachmeier: Weird formatting here. Remove the "-" and just make this a sentence after "are exposed to it." above

docs/source/tutorials/risk.rst
```diff
@@ -0,0 +1,282 @@
...
+
+This tutorial focuses on the exposure side: how the
+:class:`~vivarium_public_health.risks.base_risk.Risk` component determines
+*who* is exposed.
+
+
+Common Setup
+------------
+
+Every code example in this tutorial uses imports and helpers shown below.
```
stevebachmeier:
```suggestion
Every code example in this tutorial uses the imports and helpers shown below.
```

docs/source/tutorials/risk.rst
```diff
@@ -0,0 +1,282 @@
...
+       ],
+       configuration=config,
+       plugin_configuration=base_plugins,
+   )
+
+   pop = sim.get_population(["test_risk.exposure"])
+   n_exposed = (pop["test_risk.exposure"] == "exposed").sum()
+   proportion = n_exposed / len(pop)
+   # Approximately 40% should be exposed.
+   print(f"Proportion exposed near 0.4: {0.35 < proportion < 0.45}")
```
stevebachmeier: This isn't very reassuring. Maybe use `np.isclose` or something like that? Or at least manually tighten the bounds

stevebachmeier: Or even better would be just to print `proportion`. Unless that's not reproducible (it should be...we do this kind of thing all the time in doctests)

docs/source/tutorials/risk_effect.rst
```diff
@@ -0,0 +1,614 @@
...
+   from vivarium_public_health.disease import *
+   from vivarium_public_health.population import BasePopulation
+   from vivarium_public_health._example_data import *
+   base_plugins = BASE_PLUGINS
+
+
+Overview
+--------
+
+A risk effect modifies a *target* rate based on exposure. A *target* is a
```
stevebachmeier:
```suggestion
A risk effect modifies a *target* rate based on exposure values of that risk. A *target* is a
```

docs/source/tutorials/risk_effect.rst
```diff
@@ -0,0 +1,614 @@
...
+specific rate modified by the risk, identified by an entity and a measure
```
stevebachmeier: clarify this. You mention a string of entity and measure, but this also includes "cause". Be sure to say which is the entity and which is the measure.

docs/source/tutorials/risk_effect.rst
```diff
@@ -0,0 +1,614 @@
...
+   unexposed = pop[pop["test_risk.exposure"] == "unexposed"]
+
+   exposed_infection_rate = (exposed["test_cause"] == "test_cause").sum() / len(exposed)
+   unexposed_infection_rate = (unexposed["test_cause"] == "test_cause").sum() / len(
+       unexposed
+   )
+
+   # With RR=5, the ratio of infection rates should be approximately 5.
+   ratio = exposed_infection_rate / unexposed_infection_rate
+   print(f"Rate ratio near 5: {3.0 < ratio < 7.0}")
```
stevebachmeier: same comment as above. I would never personally view 3-7 as "near" 5. Just print the ratio and show that it's "near" 5

docs/source/tutorials/risk_effect.rst
```diff
@@ -0,0 +1,614 @@
...
+   pollution_only_rate = (pollution_only["test_cause"] == "test_cause").sum() / len(
+       pollution_only
+   )
+   neither_rate = (neither_exposed["test_cause"] == "test_cause").sum() / len(
+       neither_exposed
+   )
+
+   # Combined RR is 3*2=6, so both-exposed vs neither should have ratio near 6.
+   both_ratio = both_rate / neither_rate
+   print(f"Both-exposed ratio near 6: {4.5 < both_ratio < 7.5}")
```
stevebachmeier: Same as above

stevebachmeier: In addition to printing to see as a human that it's "close" to the value, you can `assert np.isclose(...)`

stevebachmeier (approved):

albrja (approved):

docs/source/tutorials/risk.rst
hussain-jafari: Updated using np.iclose and tighter bounds for all instances.

PR #651, opened 2026-05-06
Support additive causal effects

rmudambi:
### Description

- *Category*: bugfix/feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7006

### Changes and notes
- New AdditiveEffect class alongside MultiplicativeEffect, driven by abstract effect_type / calibration_constant_type properties on CausalFactorEffect.
- RiskEffect and InterventionEffect now subclass MultiplicativeEffect.
- Renamed relative_risk_* → effect_* across CausalFactorEffect, with a _deprecated_method_shim decorator preserving the old names with DeprecationWarning.
- Effect-type-aware combiner/reduction. Multiplicative uses multiplication_combiner + 1 - raw_union(c_i); additive uses addition_combiner + -sum(c_i). Both preserve the population-level baseline.
- Documentation overhaul of calibration_constant.py

Naming improvements:
- Static get_name(...) on RiskEffect/InterventionEffect, Risk.risk/Intervention.intervention properties
- Static DichotomousDistribution.get_exposed/get_unexposed/rename_deprecated_categories.

Bug fixes:
- Properly-indexed correct_target_mask in get_filtered_data
- Empty-config check in CausalFactor.get_distribution_type.

### Testing
Ran make check
Added new tests
Ran with MNCH NO Child model

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,10 @@
+**5.1.8 - 05/08/26**
+
+- Feature: Enable additive effects in addition to multiplicative effects 
+- Refactor: Rename internal CausalFactorEffect methods for clarity
+- Feature: enable risk-affected pipelines to modify multi-column attribute pipelines
+- Minor bugfixes
+
 **5.1.7 - 04/30/26**
 
 - Add disease tutorial
--- a/src/vivarium_public_health/causal_factor/calibration_constant.py
+++ b/src/vivarium_public_health/causal_factor/calibration_constant.py
@@ -62,8 +62,13 @@ def register_risk_affected_attribute_producer(
         in addition to the calibration constant post-processor. These will be
         applied after the calibration constant post-processor.
     """
+    post_processors = (
+        additional_post_processors
+        if isinstance(additional_post_processors, Sequence)
+        else [additional_post_processors]
+    )
     _RiskAffectedPipeline.create(
-        builder, name, source, required_resources, additional_post_processors, is_rate=False
+        builder, name, source, required_resources, post_processors, is_rate=False
     )
 
 
@@ -96,8 +101,13 @@ def register_risk_affected_rate_producer(
         in addition to the calibration constant post-processor. These will be
         applied after the calibration constant post-processor.
     """
+    post_processors = (
+        additional_post_processors
+        if isinstance(additional_post_processors, Sequence)
+        else [additional_post_processors]
+    )
     _RiskAffectedPipeline.create(
-        builder, name, source, required_resources, additional_post_processors, is_rate=True
+        builder, name, source, required_resources, post_processors, is_rate=True
     )
 
 
@@ -111,7 +121,7 @@ class _RiskAffectedPipeline(Component):
         name: str,
         source: Callable[..., pd.Series],
         required_resources: Sequence[str],
-        additional_post_processors: AttributePostProcessor | Sequence[AttributePostProcessor],
+        additional_post_processors: Sequence[AttributePostProcessor],
         is_rate: bool,
     ) -> None:
         """Factory method to create and set up the class."""
@@ -124,7 +134,7 @@ class _RiskAffectedPipeline(Component):
         target_pipeline_name: str,
         target_pipeline_source: Callable[..., pd.Series],
         required_resources: Sequence[str | Resource],
-        additional_post_processors: AttributePostProcessor | Sequence[AttributePostProcessor],
+        additional_post_processors: Sequence[AttributePostProcessor],
         is_rate: bool,
     ):
         super().__init__()
@@ -158,12 +168,6 @@ class _RiskAffectedPipeline(Component):
             else builder.value.register_attribute_producer
         )
 
-        preferred_post_processor_list = (
-            list(self._additional_post_processors)
-            if isinstance(self._additional_post_processors, Sequence)
-            else [self._additional_post_processors]
-        )
-
         register_pipeline(
             self._target_pipeline_name,
             source=self._target_pipeline_source,
@@ -171,7 +175,7 @@ class _RiskAffectedPipeline(Component):
             preferred_combiner=multiplication_combiner,
             preferred_post_processor=[
                 self._apply_calibration_constant,
-                *preferred_post_processor_list,
+                *self._additional_post_processors,
             ],
         )
 
--- a/src/vivarium_public_health/causal_factor/calibration_constant.py
+++ b/src/vivarium_public_health/causal_factor/calibration_constant.py
@@ -10,7 +10,7 @@ constants.
 
 """
 from collections.abc import Callable, Sequence
-from typing import Any
+from typing import Any, Literal
 from typing import SupportsFloat as Numeric
 
 import pandas as pd
@@ -22,6 +22,7 @@ from vivarium.framework.resource import Resource
 from vivarium.framework.values import (
     AttributePostProcessor,
     ValuesManager,
+    addition_combiner,
     multiplication_combiner,
     raw_union_post_processor,
 )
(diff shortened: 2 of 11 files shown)


rmudambi:
## PR Review: Refine Calibration Constants — Enable Additive Effects & Method Renames

### Summary

This PR (branch `rmudambi/mic-7006/refine-calibration-constants`) extends the calibration constant infrastructure with three major features: (1) adding an `effect_type` parameter to `register_risk_affected_*_producer` supporting both multiplicative and additive pipeline effects, (2) introducing a `CausalFactorEffect` ABC hierarchy with `MultiplicativeEffect` and `AdditiveEffect` subclasses in a new `causal_factor/effect.py`, and (3) renaming RR-specific method names to generic effect terminology (with deprecation shims). It also removes a duplicate `risks/calibration_constant.py` file from a failed merge and cleans up duplicate imports in disease modules.

The calibration constant application mechanism has changed from a `preferred_post_processor` (which explicitly guarded against zero-valued entries) to a `register_attribute_modifier` approach (which participates in the combiner chain alongside causal factor effect modifiers).

### Design

1. **Missing zero-value guard for additive effects** — calibration_constant.py: The old `_apply_calibration_constant` post-processor explicitly only applied to non-zero entries (`non_zero_index = value[value != 0].index`). The new approach registers the calibration constant table as an unconditional `attribute_modifier`. For multiplicative effects this is fine (`0 × anything = 0`), but for additive effects with `addition_combiner`, a zero base value becomes `0 + (-calibration) = -calibration`, producing nonsensical negative values. Consider adding a guarded modifier or documenting that calibration data must be zero wherever the base rate is zero.

2. **Deprecation shims give deprecated method precedence over new overrides** — effect.py: If a subclass overrides *both* the deprecated method (e.g., `build_rr_lookup_table`) and the new one (`build_effect_lookup_table`), the `hasattr` check will always find the deprecated method first and use it, silently ignoring the new override. Consider checking whether the new method was overridden on the concrete class before falling back to the deprecated path.

3. **`sum(value)` for additive effects relies on implicit type promotion** — calibration_constant.py: `value` starts as `[0]` (int) and grows with appended `pd.Series`. Python's `sum([0, series1, series2])` works via `((0 + series1) + series2)` but is brittle if the seed ever changes type. Consider explicit accumulation mirroring the multiplicative path's use of `raw_union_post_processor`.

4. **Factory `create()` returns `None`** — calibration_constant.py: Minor ergonomic limitation. Returning the instance would allow tests/callers to inspect the component without searching the builder registry.

### Maintainability

1. **Unexplained negation in additive post-processor** — calibration_constant.py: `joint_calibration_constant = -sum(value)` — the sign flip is semantically critical but undocumented. Add an inline comment explaining that the negation counteracts the additive effects so the population-level baseline is preserved.

2. **`register_attribute_modifier` semantics undocumented** — calibration_constant.py: The switch from `preferred_post_processor` to `register_attribute_modifier` changes when/how the calibration constant participates in the pipeline. Add a comment explaining the design decision (modifier participates in the combiner chain alongside effect modifiers).

3. **Implicit coupling between `CausalFactorEffect` and `DichotomousDistribution`** — effect.py: `process_categorical_data` calls `DichotomousDistribution.get_exposed(causal_factor_type)` directly. Document this cross-module dependency in the method docstring.

### DRY

1. **Repeated deprecation shim pattern (×6)** — effect.py and five other locations: The `hasattr`/`warnings.warn`/`return` boilerplate is repeated for `build_rr_lookup_table`, `load_relative_risk`, `rebin_relative_risk_data`, `_rebin_relative_risk_data`, `get_relative_risk_source`, and `register_relative_risk_pipeline`. Extract a `_deprecated_method_shim` decorator to centralize the pattern:

    ```python
    def _deprecated_method_shim(old_name: str):
        def decorator(method):
            @functools.wraps(method)
            def wrapper(self, *args, **kwargs):
                if hasattr(self, old_name):
                    warnings.warn(
                        f"{self.__class__.__name__} defines `{old_name}`, deprecated. "
                        f"Rename to `{method.__name__}`.", DeprecationWarning, stacklevel=2)
                    return getattr(self, old_name)(*args, **kwargs)
                return method(self, *args, **kwargs)
            return wrapper
        return decorator
    ```

2. **Near-identical registration functions** — calibration_constant.py: `register_risk_affected_attribute_producer` and `register_risk_affected_rate_producer` differ only in `is_rate`. Extract a shared `_register_risk_affected_producer` helper and have both delegate to it.

### Tests

1. **No test for `calibration_constant=1` with additive effects** — test_calibration_constant.py: `test_calibration_constant_of_one_zeroes` is not parameterized over `effect_type`. For multiplicative `c=1` zeros the output, but for additive it should yield `base - 1`. This boundary case should be explicitly tested per effect type.

2. **No edge-case tests for calibration > 1 or negative calibration** — test_calibration_constant.py: All test values are in `(0, 1)`. If calibration > 1 is invalid, the code should reject it; if valid, the output (potentially negative) should be tested.

3. **No test for zero base value** — test_calibration_constant.py: With `base_value=0` and additive effects, the output would be `-calibration`. This edge case is untested and relates to design concern #1 above.

4. **Hard-coded date assertion in deprecation test** — test_effect.py: `assert datetime.date.today() <= datetime.date(2026, 8, 6)` will cause CI failures after the deadline for anyone touching the repo. Consider a less disruptive pattern (e.g., a dedicated `test_deprecation_deadlines.py` file or a `TODO` with a lint rule).

5. **Deprecation test only verifies the deprecated path, not the new path** — test_effect.py: The test proves old methods dispatch correctly but doesn't verify that a vanilla `RiskEffect` (without old methods) executes the new logic without warnings.

### Documentation

1. **Module docstring references "RiskEffect components"** — calibration_constant.py: Should say "CausalFactorEffect components" to match the function docstrings.

2. **`post_processors` param docstring says "attribute pipeline" in rate function** — calibration_constant.py: Should say "rate pipeline."

3. **`register_target_modifier` docstring says "Register the relative risk"** — effect.py: Should generalize to "Register the effect as a modifier on the target pipeline."

4. **CHANGELOG incomplete** — CHANGELOG.rst: Only mentions post-processors. Should also document `effect_type` parameter, `AdditiveEffect`/`MultiplicativeEffect` classes, and method renames.

5. **`AdditiveEffect` lacks usage guidance** — effect.py: Docstring doesn't explain what data shape `additive_effect` expects or how the calibration constant is computed for additive pipelines.

6. **Extra blank line in `_calibration_constant_post_processor` docstring** — calibration_constant.py: Trailing blank line before closing `"""`.

### Functionality

1. **Behavioral change: calibration constant now applies to zero-valued entries** — The old `_apply_calibration_constant` explicitly skipped zero-valued entries. The new `register_attribute_modifier` approach applies unconditionally. For multiplicative effects this is algebraically safe, but for additive effects it can produce negative pipeline values. This is a semantic regression for any existing caller that relied on the zero-guarding behavior.

2. **Additive calibration constant sign convention requires careful data alignment** — The additive calibration constant is computed as `-sum(modifier_values)` and then added to the pipeline via `addition_combiner`. This means the raw modifier values should be *positive* numbers representing the amount each effect *contributes* to the pipeline value. If a user provides a PAF-like fraction for an additive effect, they'll get incorrect results. The API doesn't enforce or document this distinction.

3. **Type annotation broadening** — calibration_constant.py: `required_resources` is typed as `Sequence[str | Resource]` (broadened from the old `Sequence[str]`). This is intentional and correct.

### Minor Nits

1. The test file header comment says `Tests for :mod:\`vivarium_public_health.risks.calibration_constant\`` but the module has been moved to `vivarium_public_health.causal_factor.calibration_constant`.
2. calibration_constant.py: `addition_combiner` is imported without any comment about its relationship to the new additive feature.

### Overall

The PR introduces a well-structured generalization from multiplicative-only effects to multiplicative/additive, with a clean class hierarchy. The main concerns are: (1) the loss of zero-value guarding for additive effects, which could produce negative pipeline values; (2) the deprecation shim pattern giving precedence to old methods even when new ones are overridden; and (3) substantial DRY violations in the deprecation boilerplate. The CHANGELOG and some docstrings need updating to reflect the full scope of changes. Test coverage should be expanded to include additive boundary cases (calibration=1, zero base values).

PR #652, opened 2026-05-22
treatment tutorials

hussain-jafari:
treatment tutorials


docs/source/tutorials/treatment.rst
```diff
@@ -0,0 +1,645 @@
...
+   from vivarium_public_health._example_data import *
+   base_plugins = BASE_PLUGINS
+
+
+Overview
+--------
+
+The treatment module provides several components for modeling interventions:
+
+**Intervention** - a dichotomous coverage model that assigns each simulant
```
albrja, line 40: I just think it is a bit confusing when the module is named "treatment" but then everything is immediately about intervention so maybe there needs to be a name change somewhere.

albrja (approved):

docs/source/tutorials/treatment.rst
```diff
@@ -0,0 +1,645 @@
+=========
+Treatment
+=========
+
+:mod:`vivarium_public_health` provides components for modeling treatment
+interventions in public health simulations. This tutorial covers the
+treatment module - how to model intervention coverage, how interventions
```
rmudambi: Isn't this a package, not a module?

rmudambi (approved):

PR #654, opened 2026-05-29
Albrja/feature/mic-6789/error-empty-rr-df

albrja:
### Runtime error if filtered relative risk dataframe is empty
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6789

### Changes and notes
-raise error if filtered relative risk dataframe is empty

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+**5.1.10 - 06/01/26**
+
+- Raise error if filtered relative risk data is empty
+
 **5.1.9 - 05/22/26**
 
 - Add new intersphinx mappings for vivarium-engine and vivarium-config-tree
--- a/src/vivarium_public_health/risks/effect.py
+++ b/src/vivarium_public_health/risks/effect.py
@@ -226,6 +226,8 @@ class NonLogLinearRiskEffect(RiskEffect):
         ------
         MissingDataError
             If the TMRED data uses draw-level TMRELs or is not found.
+        ValueError
+            If the filtered relative risk data is empty.
         """
         if configuration is None:
             configuration = self.configuration
@@ -248,6 +250,14 @@ class NonLogLinearRiskEffect(RiskEffect):
         # calculate RR at TMREL
         rr_source = configuration.data_sources.relative_risk
         original_rrs = self.get_filtered_data(builder, rr_source)
+
+        if isinstance(original_rrs, pd.DataFrame) and original_rrs.empty:
+            raise ValueError(
+                f"The relative risk data for {self.causal_factor.name} affecting "
+                f"{self.target.name} {self.target.measure} is empty. This can happen "
+                "when the data contains no rows matching the affected entity and "
+                "measure of this risk effect."
+            )
 
         self.validate_rr_data(original_rrs)
 
--- a/src/vivarium_public_health/risks/effect.py
+++ b/src/vivarium_public_health/risks/effect.py
@@ -227,7 +227,9 @@ class NonLogLinearRiskEffect(RiskEffect):
         MissingDataError
             If the TMRED data uses draw-level TMRELs or is not found.
         ValueError
-            If the filtered relative risk data is empty.
+            If the relative risk data fails validation (e.g. it is empty,
+            or its ``parameter`` column is non-numeric or not monotonically
+            increasing). See :meth:`validate_rr_data`.
         """
         if configuration is None:
             configuration = self.configuration
@@ -250,14 +252,6 @@ class NonLogLinearRiskEffect(RiskEffect):
         # calculate RR at TMREL
         rr_source = configuration.data_sources.relative_risk
         original_rrs = self.get_filtered_data(builder, rr_source)
-
-        if isinstance(original_rrs, pd.DataFrame) and original_rrs.empty:
-            raise ValueError(
-                f"The relative risk data for {self.causal_factor.name} affecting "
-                f"{self.target.name} {self.target.measure} is empty. This can happen "
-                "when the data contains no rows matching the affected entity and "
-                "measure of this risk effect."
-            )
 
         self.validate_rr_data(original_rrs)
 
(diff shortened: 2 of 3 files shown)


albrja:
Code Reiewer
PR Review: Prevent loading empty relative risk dataframe (#654)
Summary
This PR adds a guard in [NonLogLinearRiskEffect.load_relative_risk](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_public_health/risks/effect.py#L254-L260) that raises a ValueError with a clear, contextual message when the filtered relative-risk data is an empty DataFrame — instead of letting the simulation fail confusingly further downstream when building the lookup table. It updates the method's docstring Raises section, adds a CHANGELOG entry (5.1.10), and adds a test that triggers the empty case via an affected_entity mismatch.

The change is small, well-targeted, and a genuine UX improvement. I confirmed functionally (running the validation logic on a filtered-to-empty frame) that validate_rr_data silently passes an empty frame — the numeric parameter dtype is preserved through filtering and groupby(...).all() on an empty frame returns True. So this guard closes a real gap rather than duplicating existing protection. The main discussion point is where the check belongs.

Design
Move the empty check into validate_rr_data (effect.py:254-262 → effect.py:336). (flagged by Design + DRY) validate_rr_data is the designated validator for rr_data and already raises ValueError for non-numeric/non-monotonic data. validate_rr_data is called from two sites — load_relative_risk:262 and build_rr_lookup_table:161 — but the new empty check protects only the first. Relocating it as the first guard in validate_rr_data covers both paths and closes the silent-pass-on-empty hole I confirmed above. The added block in load_relative_risk can then be deleted (the existing self.validate_rr_data(original_rrs) call covers it):


def validate_rr_data(self, rr_data: pd.DataFrame) -> None:
    if rr_data.empty:
        raise ValueError(
            f"The relative risk data for {self.causal_factor.name} affecting "
            f"{self.target.name} {self.target.measure} is empty. ..."
        )
    ...
Since validate_rr_data is already typed (rr_data: pd.DataFrame), the isinstance(...) guard becomes unnecessary there.

isinstance(original_rrs, pd.DataFrame) guard silently skips the scalar path (effect.py:254). (flagged by Design + Maintainability) get_filtered_data is typed float | pd.DataFrame, but everything after line 264 (.columns, .groupby, .merge) treats original_rrs as a DataFrame and would crash with AttributeError on a scalar. The guard thus protects against a case the rest of the method doesn't support, implying scalar input is a handled path when it isn't. If the check stays in load_relative_risk, consider asserting the contract explicitly; if it moves to validate_rr_data (finding 1), the guard disappears entirely.

ValueError is the right exception (no change needed). It's consistent with the other rr-data validation failures in validate_rr_data; MissingDataError is reserved in this module for "data not modeled / not found" (TMRED cases), not structurally-unusable loaded data.

Maintainability
Error message overstates a single cause (effect.py:255-260). (flagged by Maintainability + Design + Documentation) The message asserts emptiness happens "when the data contains no rows matching the affected entity and measure." Per get_filtered_data, filtering on affected_entity/affected_measure only happens if those columns exist — the frame could also be empty because the source data was empty to begin with. Soften to avoid sending debuggers down the wrong path:

"...is empty. This can happen if the source data is empty, or contains no rows matching this effect's affected entity ({self.target.name}) and measure ({self.target.measure})."

(If the check stays put) add a one-line comment explaining the isinstance guard exists only to satisfy the broad get_filtered_data return type, and that the empty-DataFrame case is the only one guarded here.

DRY
Same as Design #1 — the empty check is a third rr-data validation living outside the validate_rr_data method that exists to centralize exactly that. No cross-module duplication exists (other .empty checks in the codebase operate on unrelated demographic/population frames), so finding #1 is the only consolidation opportunity.
Tests
match="empty" is too weak (test_effect.py:643). The word "empty" could appear in an unrelated future ValueError (or one raised by pandas/lookup machinery), so the test wouldn't catch removal of the new check. Assert a distinctive fragment of the actual message:

with pytest.raises(ValueError, match="relative risk data for test_risk affecting test_cause incidence_rate is empty"):
No coverage for an affected_measure mismatch (test_effect.py:612-644). The test only covers entity-mismatch. A measure-mismatch (affected_entity="test_cause" but affected_measure="excess_mortality_rate") is an equally plausible misconfiguration that flows through a distinct masking branch. Add a parametrized case.
No coverage for an already-empty input DataFrame (zero rows, correct columns) passed directly — the canonical degenerate input, exercising the original_rrs.empty half independent of filtering logic. Cheap to add.
Good: the test correctly exercises the real filtering path (get_filtered_data producing the empty frame) rather than injecting a pre-empty frame, and the happy path is already covered by test_non_loglinear_effect.
Documentation
CHANGELOG date is future-dated (CHANGELOG.rst:1). 06/01/26 is 3 days ahead of today (2026-05-29). Use the actual merge/release date to keep the changelog accurate.
CHANGELOG entry underspecifies scope (CHANGELOG.rst:3). The check is specific to NonLogLinearRiskEffect; "Raise error if filtered relative risk data is empty" reads as if all risk effects now do this. Scope it: - Raise a ValueError in NonLogLinearRiskEffect when the filtered relative risk data is empty.
Raises section now incomplete (effect.py:229-230). Now that the section enumerates one ValueError cause, the omission of the pre-existing validate_rr_data ValueErrors (non-numeric / non-monotonic parameter) is more misleading. Expand the ValueError entry to cover all cases. (If you adopt Design #1, keep this entry since the cause remains reachable via validate_rr_data.)
(Optional) Append the configured data source (configuration.data_sources.relative_risk, available at line 251) to the message to speed debugging of wrong-artifact/override cases.
Functionality
The guard is necessary and correctly ordered (verified). I ran the validate_rr_data logic against a filtered-to-empty frame: the numeric parameter dtype survives filtering, so the numeric check passes, and groupby(demographic_cols).apply(...).all() returns True on the empty frame — validate_rr_data would not catch it. Without this PR, the empty frame flows into groupby(...).apply(get_rr_at_tmrel) and the merge, producing a confusing downstream failure. The new check correctly intercepts before that.
No regression on the scalar path. When get_filtered_data returns a scalar, the guard is skipped — but for NonLogLinearRiskEffect a scalar already fails downstream regardless, so behavior is unchanged (see Design #2 for the cleaner contract).
build_rr_lookup_table path is covered because it calls load_relative_risk (which now contains the check). Moving the check to validate_rr_data per Design #1 would preserve this and additionally protect any direct validate_rr_data callers.
Minor Nits
CHANGELOG date format/future-dating (Documentation #1).
Message could name the data source (Documentation #4).
Overall
A small, correct, and worthwhile fix that converts a confusing downstream failure into a clear, actionable error — and I verified it closes a genuine gap (validate_rr_data silently accepts empty frames). The single highest-value improvement, raised independently by the design and DRY lenses, is to relocate the check into validate_rr_data so both call sites are protected and rr-data validation stays centralized; that also lets the isinstance guard go away. Secondary priorities: tighten the test's match assertion and add the measure-mismatch / already-empty cases, soften the error message's single-cause wording, fix the future-dated changelog, and complete the docstring Raises section. None of these are blockers.

rmudambi (approved):

hussain-jafari (approved):

